PR #120, opened 2020-03-29
Links are not broken anymore when node titles are renamed

daviddq:
This commit solves 2 related things:

1) When saving a node, the spaces on links are trimmed from right and left for consistency

2) When renaming a node, the body links of the nodes' pointing to the renamed node are updated so links are not broken

PR #121, opened 2020-03-30
Parent links are no longer broken when changing node title

daviddq:
Fixes issues:

- https://github.com/YarnSpinnerTool/YarnEditor/issues/122
- https://github.com/YarnSpinnerTool/YarnEditor/issues/123

### DIFF
--- a/src/js/classes/app.js
+++ b/src/js/classes/app.js
@@ -1085,9 +1085,13 @@ export var App = function(name, version) {
   };
 
   this.editNode = function(node) {
-    if (node.active()) {
-      // console.log(node)
-      self.editing(node);
+    if ( !node.active() ) {
+      return;
+    }
+
+    node.oldTitle = node.title(); // To check later if "title" changed
+
+    self.editing(node);
 
       $('.node-editor')
         .css({ opacity: 0 })
@@ -1234,7 +1238,6 @@ export var App = function(name, version) {
 
       self.toggleSpellCheck();
       self.updateEditorStats();
-    }
   };
 
   this.chooseRelativePathImage = function(imagePath) {
@@ -1517,9 +1520,12 @@ export var App = function(name, version) {
 
   this.saveNode = function() {
     if (self.editing() != null) {
+      self.editing().title ( document.getElementById('editorTitle').value.trim() );
+      self.editing().body ( self.trimBodyLinks( self.editing().body() ) );
+
       self.makeNewNodesFromLinks();
-      self.updateNodeLinks();
-      self.editing().title(self.trim(self.editing().title()));
+      self.updateNodeLinksStartingFromNode(self.editing());
+
       $('.node-editor').transition({ opacity: 0 }, 250);
       $('.node-editor .form').transition({ y: '-100' }, 250, function(e) {
         self.editing(null);
@@ -1587,8 +1593,43 @@ export var App = function(name, version) {
     }
   };
 
+  this.trimBodyLinks = function ( body ) {
+    var re = /\[\[(.+?)\|\s*(.+?)\s*\]\]/g
+    return body.replace(re, '[[$1\|$2]]');
+  };
+
   this.updateNodeLinks = function() {
     for (var i in self.nodes()) self.nodes()[i].updateLinks();
+  };
+
+  // TODO: probably 'updateNodeLinksStartingFromNode' can be used as a
+  // replacement for 'updateNodeLinks'. I'll check it in next iterations.
+  this.updateNodeLinksStartingFromNode = function(node) {
+    var toUpdate = [];
+    var updated = [];
+
+    toUpdate.push(node);
+
+    while (node = toUpdate.pop() ) {
+      if ( updated.includes(node) ) {
+        continue;
+      }
+
+      updated.push( node );
+
+      node.updateLinks();
+
+      node.linkedTo().forEach( child => {
+        if ( !updated.includes ( child ) ) {
+          toUpdate.push( child );
+        }
+      });
+
+      node.linkedFrom().forEach( parent => {
+        if ( !updated.includes ( parent ) )
+          toUpdate.push( parent );
+      });
+    }
   };
 
   this.makeNewNodesFromLinks = function() {
--- a/src/js/classes/app.js
+++ b/src/js/classes/app.js
@@ -1594,7 +1594,7 @@ export var App = function(name, version) {
   };
 
   this.trimBodyLinks = function ( body ) {
-    var re = /\[\[(.+?)\|\s*(.+?)\s*\]\]/g
+    var re = /\[\[(.+?)\|\s*(.+?)\s*\]\]/g;
     return body.replace(re, '[[$1\|$2]]');
   };
 
--- a/src/js/classes/app.js
+++ b/src/js/classes/app.js
@@ -1085,7 +1085,7 @@ export var App = function(name, version) {
   };
 
   this.editNode = function(node) {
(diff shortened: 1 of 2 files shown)


blurymind:
thank you for this pr, I'll have a careful look tonight.
Can you please put links to the issue tickets that this pr fixes? : )

daviddq:
This PR fixes issues:

- https://github.com/YarnSpinnerTool/YarnEditor/issues/122
- https://github.com/YarnSpinnerTool/YarnEditor/issues/123

daviddq:
On PR #120 @blurymind mentioned:

> The title trimming is a genuine bug, but the other thing - changing a link to a node disconnects it from the old node and creates a new one if that new one doesnt exists - thats by design.

I assumed that was a bug because:

1) If it's a feature it behaves incorrectly since arrows are not immediately updated. You need to enter a node and save for the arrows to get updated.

2) If it's a feature it behaves incorrectly since the new node is not immediately created. You need to enter a node and save for the new node to be created.

3) My proposal mimics how Twine 2 handles the title change use case. In my humble opinion makes more sense this way and helps keeping consistency and not breaking things inadvertently.

blurymind:
Ok, let me try it tonight :) thank you for working on this

blurymind:
Will need to do better testing on this. I was able to get it in an odd state when making copies  of nodes (select node, ctrl+c, ctrl+v)

daviddq:
I didn't even know that copy/pasting worked on nodes. I'll check that this afternoon. If you can describe "steps to reproduce" would be super nice. Thanks for your time.

blurymind:
Just kept making copies of nodes to get it to have two nodes with the same name, then editing a node that links to two nodes with the same name and changing the link name to see what will happen.
Something odd happens when you delete one of the copies. I need some time to give exacts steps.

Other than that I notices that if you delete a node that has a link to it, the link to it gets appended with an "err" for some reason

daviddq:
The copy/paste functionality is already rotten on _master_ the moment Yarn Editor allows two nodes to exist with the same title since the title is used as a node id AFAIK (correct me if I'm wrong). You can notice on _master_ that if you create some nodes with the same title you end up unable to move/edit/etc some nodes since they become unreachable for the editor.

I would like to help as much as possible to fine tune this bugs, so I'm all ears to your feedback.

blurymind:
Yes indeed we need to fix this. Can you create another PR that addresses this. Ideally we want it to append to copied nodes "copyOf" + theNodeTheyCopiedTitle

src/js/classes/app.js
```diff
@@ -1587,10 +1593,45 @@ export var App = function(name, version) {
     }
   };
 
+  this.trimBodyLinks = function ( body ) {
```
blurymind: this is a nitpick, but I notice you add an empty space around (), which is inconsistent with the rest. Do you by any chance use prettier?

daviddq: I can fix it this. I have no strong preference for one notation or another at all. I completely disabled my auto formatter since the rules I use are very different from the ones used by the project (number of indentation spaces, courly brackets, etc.)

src/js/classes/node.js
```diff
@@ -372,9 +372,10 @@ export var Node = function() {
     return false;
   };
 
-  this.getLinksInNode = function() {
-    // find all the links
-    var links = self.body().match(/\[\[(.*?)\]\]/g);
+  this.getLinksInNode = function(node) {
+    node = node || self;
+
+    var links = node.body().match(/\[\[(.*?)\]\]/g);
```
blurymind: instead of overwriting a parameter value, which is not good practice, you can do
`var links = (node || self).body().match(/\[\[(.*?)\]\]/g);
`
here

src/js/classes/app.js
```diff
@@ -1085,9 +1085,13 @@ export var App = function(name, version) {
   };
 
   this.editNode = function(node) {
-    if (node.active()) {
-      // console.log(node)
-      self.editing(node);
+    if ( !node.active() ) {
+      return;
+    }
```
blurymind: :+1:

src/js/classes/app.js
```diff
@@ -1587,10 +1593,45 @@ export var App = function(name, version) {
...
 
+  // TODO: probably 'updateNodeLinksStartingFromNode' can be used as a
+  // replacement for 'updateNodeLinks'. I'll check it in next iterations.
+  this.updateNodeLinksStartingFromNode = function(node) {
+    var toUpdate = [];
+    var updated = [];
+
+    toUpdate.push(node);
+
+    while (node = toUpdate.pop() ) {
```
blurymind: careful with while loops :) can this ever be forever true?

src/js/classes/node.js
```diff
@@ -398,24 +399,50 @@ export var Node = function() {
...
+  this.updateLinksFromParents = function() {
+    // If title didn't change there's nothing we need to update on parents
+    if ( !self.oldTitle || (self.oldTitle === self.title()) ) {
+      return;
+    }
+
+    self.linkedFrom.removeAll();
+
+    app.nodes().forEach( parent => {
+      var parentLinks = self.getLinksInNode(parent);
```
blurymind: this is really cool. I really like that you made a handy function to collect the links

blurymind:
I had another play with this and it seems pretty solid, you've done a pretty good job here. The bugs I encountered before might likely be in master anyway.  I left a few notes on the code, but nothing major :) 

Some parts seemed a bit risky - particularly the one where you find links in the body and overwrite them. But I guess we kind of need that anyways, we just need to test for regressions really carefully

daviddq:
Cool! I can easily fix the points you menction on the PR review. For that I need to submit a new PR, right?

blurymind:
> Cool! I can easily fix the points you menction on the PR review. For that I need to submit a new PR, right?

No need, just push changes to the branch this one uses :)
The changes will appear on the pull request

src/js/classes/app.js
blurymind: Ah in that case please set them to be similar to this project or keep it disabled. Don't want the dif to get contaminated with some formatting that is not compatible with the rest :)

blurymind:
One thing that I noticed is that when you rename nodes, then sometimes when you use the "go back to last edited" button, Yarn creates new nodes with the old names

blurymind:
Have a look here:
ezgif-4-5f6b1e239e9a


Does this happen on master?

src/js/classes/app.js
daviddq: Regarding the "while" condition:

  - It pops a node from "toUpdate"
  - Assigns the node to "node" variable
  - The body of the while will be entered only if "node" is "truthy" not null, undefined, 0, false... briefly, if we could pop a node from "toUpdate".

Regarding the algorithm: it's a sort of flood-fill. We start updating the node we just updated. Then we update the parents and children, then the parants and childrens of the later, etc. If there are unconnected nodes, they are not updated (there's no need to). If there are cyclical references we avoid following them more than once with the "updated" array (an array of the already visited nodes). Pretty standard algorithm.

The important thing here is that we start updating nodes from the node being edited.  "updateNodeLinks()" fails to refresh the arrows because it traverses de "nodes" array from beginning to end. 

That's why I proposed on the comments to use this method instead for all node updates.

daviddq:
Yes... same behavior on _master_

daviddq:
I pushed the changes you proposed.

blurymind:
Merged :tada: Thank you for working on this.

Issue #128, opened 2020-03-31
Copy/Paste  is broken in master


daviddq:
**Version: 0.4.1**

**Current behavior**
When copying a node with ctrl+c ctrl+v the resulting node seems to be referencing the original. My bet is that the new object is not deep-cloned and some/all references to the original node are kept. Also probably related with knockout observables.

![copyPasteBroken](https://user-images.githubusercontent.com/6022674/78074365-ee695c80-73a2-11ea-90b5-f55d485aab45.gif)

**Expected behavior**
The new nodes are completely independent from the original ones.

daviddq:
I have a clearer idea on what's happening. On copy/pasting nodes, the clones are being cloned on "copy" instead of on "paste". This way, if you "copy" one and "paste" many, nodes are cloned one time and then all "pastes" make reference to the same clone.

blurymind:
Good catch! Feel free to submit the pr :)

daviddq:
Submited PR #130

blurymind:
merged now! Thank you for fixing it :)

PR #130, opened 2020-04-01
This PR fixes Issue #128 (copy paste is broken in master)

daviddq:
This PR fixes Issue #128 (copy paste is broken in master)

### DIFF
--- a/src/js/classes/app.js
+++ b/src/js/classes/app.js
@@ -478,17 +478,9 @@ export var App = function(name, version) {
         if ((e.metaKey || e.ctrlKey) && e.keyCode == 65) {
           self.selectAllNodes();
         }
+        // ctrl + v NODES
         if ((e.metaKey || e.ctrlKey) && e.keyCode == 86) {
-          // ctrl + v NODES
-          if (self.nodeClipboard.length) {
-            self.deselectAllNodes();
-            self.nodeClipboard.forEach(function(node) {
-              self.nodes.push(node);
-              self.addNodeSelected(node);
-              self.recordNodeAction('created', node);
-            });
-            self.updateNodeLinks();
-          }
+          self.pasteNodes();
         }
         // console.log(e.keyCode+"-"+e.key)
         if (e.keyCode === 46 || e.key === 'Delete') {
@@ -511,6 +503,7 @@ export var App = function(name, version) {
           self.saveNode();
         }
       }
+
       // Spacebar toggle between nodes
       if (e.keyCode === 32) {
         if (self.editing() !== null && !e.altKey) {
@@ -992,6 +985,31 @@ export var App = function(name, version) {
     }
   };
 
+  this.pasteNodes = function() {
+    if (!self.nodeClipboard.length) {
+      return;
+    }
+
+    self.deselectAllNodes();
+
+    self.nodeClipboard.forEach(function(copiedNode) {
+      var node = new Node({
+        title: self.createTitleForCopiedNode(copiedNode.title()),
+        body: copiedNode.body(),
+        tags: copiedNode.tags(),
+        colorID: copiedNode.colorID(),
+        x: copiedNode.createX,
+        y: copiedNode.createY
+      });
+
+      self.nodes.push(node);
+      self.addNodeSelected(node);
+      self.recordNodeAction('created', node);
+    });
+
+    self.updateNodeLinks();
+  }
+
   this.addNodeSelected = function(node) {
     var index = self.nodeSelection.indexOf(node);
     if (index < 0) {
@@ -1018,14 +1036,14 @@ export var App = function(name, version) {
 
   this.cloneNodeArray = function(nodeArray) {
     return nodeArray.map(function(oldNode) {
-      var node = new Node();
-      node.title(oldNode.title());
-      node.body(oldNode.body());
-      node.tags(oldNode.tags());
-      node.colorID(oldNode.colorID());
-      node.createX = oldNode.x() + 10;
-      node.createY = oldNode.y() + 10;
-      return node;
+      return new Node({
+        title: oldNode.title(),
+        body: oldNode.body(),
+        tags: oldNode.tags(),
+        colorID: oldNode.colorID(),
+        x: oldNode.x() + 10,
+        y: oldNode.y() + 10,
+      });
     });
   };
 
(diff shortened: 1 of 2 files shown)


src/js/classes/app.js
```diff
@@ -2233,4 +2266,29 @@ export var App = function(name, version) {
...
+      baseTitle = matches[1];
+      counter = Number(matches[3]);
+    }
+
+    var currentlyUsedTitles = self.getOtherNodeTitles();
+    if (!currentlyUsedTitles.includes(baseTitle)) {
+      return baseTitle;
+    }
+
+    for (;; ++counter) {
```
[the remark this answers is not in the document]
daviddq: image

Initialization and exit condition are moved out of the for statement. You coud wirte the same like:

do {
  ++counter;
  ...
} while (true);

or:

while (++counter) {
...
}

src/js/classes/node.js
```diff
@@ -18,23 +19,25 @@ export var Node = function() {
     '#C47862',
     '#97E1E9'
   ];
+
   // primary values
+  // TODO: I strongly think "this.index" is now unused
```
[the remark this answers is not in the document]
daviddq: Understood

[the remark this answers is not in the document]
daviddq: 👍

src/js/classes/app.js
```diff
@@ -1520,11 +1538,26 @@ export var App = function(name, version) {
 
   this.saveNode = function() {
     if (self.editing() != null) {
-      self.editing().title (document.getElementById('editorTitle').value.trim());
-      self.editing().body (self.trimBodyLinks(self.editing().body()));
+      // Trim spaces from the title.
+      // I use document.getElementById because self.title() still returns the
+      // old title. I would like not having to do this. Any ideas?
+      var title = document.getElementById('editorTitle').value.trim();
```
blurymind, line 1544: I believe it's because of the way its binded in the node class with the KO library, which yarn has been using for years. I personally would love to get rid of that thing in terms of having it as a dependency - but its hard to do it without having to refactor more stuff.

You can see I've done the same as you in other parts of this same file- using document.getElementById..

src/js/classes/app.js
```diff
@@ -1520,11 +1538,26 @@ export var App = function(name, version) {
...
+
+      // Make sure the new title is unique. Otherwise, put a trailing number
+      // or increment the existing one if any
+      title = self.getUniqueTitle(title);
+
+      // Update the title in the UI
+      // I use document.getElementById because self.title(title)
+      // does not immediatly change the value. I would like not having to do
+      // this. Any ideas?
+      document.getElementById('editorTitle').value = title;
```
blurymind, line 1554: no need to get editor title again, you already have it in a variable higher up

src/js/classes/app.js
```diff
@@ -2233,4 +2266,29 @@ export var App = function(name, version) {
...
   };
+
+  this.getUniqueTitle = function(desiredTitle) {
+    var baseTitle = desiredTitle || 'Node';
+    var counter = 2;
+
+    // If the title ends with "_[number]" use the same prefix with next number
+    const re = /^(.*)(_([0-9]+))$/;
+    const matches = baseTitle.match(re);
+    if (matches && matches.length===4) {
```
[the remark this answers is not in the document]
daviddq: I did, but Prettier replaces all ' by ", making the diff to explode. Maybe you have a local .editorconfig with rules to share with me.

src/js/classes/app.js
daviddq: Replacing KO is a big task. Maybe when I get more confident and confortable with the project 😅

src/js/classes/app.js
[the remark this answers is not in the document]
blurymind: ah cool beans. Thank you for explaining this :+1:

src/js/classes/app.js
blurymind: You or anyone else reading this - I will applaud the effort. This thing has caused me grief in the past, when porting yarn to be a pwa + to use webpack

blurymind:
The Pr is good to go in imo, just needs to change or cleanup the comments. Testing it for a while, it does exactly what it says on the tin and causes no regressions

src/js/classes/app.js
blurymind: The  codebase does need some spring cleaning imo. There is some dead code lying around, some ancient libraries shoved in. I might do it at some point when I get more time too :)

src/js/classes/app.js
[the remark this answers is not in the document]
blurymind: Prettier has a singleQuote option you have to enable somewhere, depending on your code editor

src/js/classes/app.js
daviddq: Line 1544 **gets** the value of DOM's title. Line 1554 **sets** the value because it's not automatically set by KO. Or do you mean getting the DOM's element reference once and use the same reference for _getting_ and _setting_?

Comments were just a call for help 😅

daviddq:
Comments are cleaned up.

src/js/classes/app.js
blurymind: yeah I was wondering why not just use the reference to set/get it :)

daviddq:
Fixed

src/js/classes/node.js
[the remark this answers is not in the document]
blurymind: lets keep it for now, as its giving each node a unique Id. This can come in handy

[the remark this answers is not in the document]
daviddq: Cool! I already deleted the comment some commits ago.

blurymind:
This looks very good now :) Thank you for doing these fixes. :+1:

blurymind:
merged :tada:

src/js/classes/app.js
blurymind: Maybe I am a bit unfair to KO, looking at its github again - it looks like it has picked up some updates in recent years

daviddq: I cannot comment since I have never used it and I do not know how it works. In all probability the problems I am having with KO are due more to my clumsiness and ignorance than to problems with the library.

blurymind: some of the code using it is very outdated too. It might be the cause for the delay

daviddq: I'm pretty confident that most of the laggy experience comes from the arrows redraw call each 25 milliseconds. I haven't had time to profile it, but smells suspicious. Arrows should only be redrawn if they need to, and only the ones that need to be redrawn.

blurymind: Someone actually complained about this on the issue tracker recently and made me wonder if some optimization can be done to make the panning experience. I thought that arrows only update after you have panned or zoomed/in or out, but not during.

I dont think we have anything to limit the redraw to only the visible arrows

blurymind: There is a function called updateArrowsThrottled, which is called when you move nodes around. I think it was also used when panning, but I disabled it then

daviddq: I would like to know what's the purpose of the `updateArrowsThrottled` function, more specifically what's the purpose of the `Utils.throttle` function on the arrows redrawing. At first sight I can't understand what it pretends to do. On my fork I reimplemented the arrows drawing with just good old SetInterval() and works like a charm.

blurymind:
In all honesty I was itching to rip it out of yarn. Its a very old function
someone wrote at some point. I sort of just moved it to utils not to break
things. It's supposed to throttle move calls, but it's confusing as to how
it does it.

daviddq:
It does nothing but to obfuscate the code. Can I rip it off? Almost all calls to updateArrows() can be done based on events and when not (during transitions) a simple setTimer() will do the trick.

blurymind:
Go ahead :) but please test for any regressions

Issue #131, opened 2020-04-01
Right click creates nodes at the same place


daviddq:
Right clicking with the mouse always creates nodes at the same place regardless the mouse pointer position.

![wrongXY](https://user-images.githubusercontent.com/6022674/78190516-6f445900-7474-11ea-8183-94ef2a3e2103.gif)

PR #132, opened 2020-04-01
Fixes issue #131 Right click creates nodes at the same place

daviddq:
Fixes issue #131 Right click creates nodes at the same place


blurymind:
Merged now! :tada:

Issue #127, opened 2020-03-31
Navigating back to previously edited nodes when the nodes have been renamed results in new nodes being created


daviddq:
**Version 0.4.1**

Navigating back to previously edited nodes when the nodes have been renamed results in new nodes being created.

![Captura de pantalla 2020-04-01 a las 8 37 37](https://user-images.githubusercontent.com/6495061/78058246-ea781300-737f-11ea-9cb1-a46b65679e83.gif)

There's a navigation stack in app.js (self.nodeVisitHistory) that keeps track of the visited nodes' titles. To avoid this problem:

A) When renaming a node, traverse self.nodeVisitHistory stack and change the affected title there too.

or

B) self.nodeVisitHistory should keep references to the nodes instead of their titles. This may result in problems when navigating back to a node that has been deleted. We should ensure that the node still exists. If not, navigate to previous.

blurymind:
Thank you for adding it to the  issue tracker :)

PR #138, opened 2020-04-05
Fixes issue #127 Navigating back to previously edited nodes when the nodes have been renamed results in new nodes being created

daviddq:
Fixes issue #127 Navigating back to previously edited nodes when the nodes have been renamed results in new nodes being created

### DIFF
--- a/src/js/classes/app.js
+++ b/src/js/classes/app.js
@@ -1538,29 +1538,20 @@ export var App = function(name, version) {
 
   this.saveNode = function() {
     if (self.editing() != null) {
-      const editorTitleElement = document.getElementById('editorTitle');
+      const editorTitleElement = $('#editorTitle')[0];
 
-      // Trim spaces from the title.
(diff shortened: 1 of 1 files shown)


blurymind:
It has some merge conflicts. Can you resolve them please :) 

You have to merge master into your branch, resolve the conflicts and then push the commit that resolves them

blurymind:
sorry about this btw, it must have happened because I recently did some fixes on the same file

daviddq:
No worries.

Conflicts solved.

blurymind:
I'm not entirely sure why, but your whole app.js file is appearing in the diff, so it is kind of impossible for me to review this properly :)

daviddq:
Same thing happened to me with your change, probably due to different prettiers or versions. Let me see...

daviddq:
I'm going to branch again from _master_ and submit a new PR since I'm not able to re-commit app.js with only my changes.

Issue #136, opened 2020-04-04
Version number inconsistencies


daviddq:
**Current behavior**
When downloading Yarn Editor the installer name on Windows is Yarn.Editor.Setup.0.2.4.exe, but when running the editor it says 0.4.1. 

Also, after latest the 4 PR the version number is still 0.4.1

**Expected behavior**
a) Installer name reflects the real version number
b) Whenever changes are introduced on the project the appropriate version segment (major, minor or patch) should be updated.

blurymind:
yes agreed :) pr welcome. Was actually wondering if we can set up some sort of an automatic electron auto building

daviddq:
I'm setting up continuous integration in my fork and I see that electron build is trying to publish to the snap store. Is this something intentional or is it just the default electron's build process?

[image](https://user-images.githubusercontent.com/6022674/78504867-eda04400-776f-11ea-9548-930d5e47a353.png)

blurymind:
I think it generates a snap package by default, which is kind of
convenient. I don't think it actually publishes it somewhere.

daviddq:
I tries to but fails. To instruct it not to publish you need to add "--publish never".

I've got the linux version automatically building and publishing the artifact on GitHub. Are you interested in CI workflows? Do you want me to PR this feature?

blurymind:
Yeah sure :) let's try it out. That way our users will get electron builds
more often too

PR #139, opened 2020-04-05
Fixes issue #136 Version number inconsistencies 

daviddq:
Fixes issue #136 Version number inconsistencies

There's still the need for someone to decide to change the version number on the main package.json


electron/main.js
```diff
@@ -2,17 +2,19 @@ const electron = require("electron");
 const ipcMain = electron.ipcMain;
 const { dialog } = electron;
 const isDev = require("electron-is").dev();
-// const fs = require("fs");
+const version = require("../package.json").version;
```
blurymind, line 5: haha clever xD never occurred to me to do just that

electron/package.json
```diff
@@ -14,12 +14,13 @@
     "url": "https://github.com/YarnSpinnerTool/YarnEditor/issues"
   },
   "scripts": {
+    "update-version": "node update-version.js",
```
blurymind, line 17: interesting, now we have a quick and easy way to update the number :D

electron/update-version.js
```diff
@@ -0,0 +1,18 @@
+var fs = require('fs')
```
blurymind, line 1: if this script updates the number of the app as well, can we move it up in the parent folder? It would make a sense to create a scripts folder at this point.

maybe 
app/scripts/update-version.js

src/js/index.js
```diff
@@ -22,9 +22,10 @@ import "jquery.transit";
 
 import "spectrum-colorpicker";
 import "lightweight-emoji-picker/dist/picker.js";
-import spoken from "spoken";
+// import spoken from "spoken";
```
blurymind, line 25: any reason this is commented out? We use spoken to transcribe voice to text and the other way around

daviddq: It was marked as "unused" by my editor. Since it's not "public/exported"-ish I thought I could safely comment it out. Am I wrong?

blurymind: Try using the transcribe feature (click on the microphone) - it will not work with this disabled.
https://twitter.com/blurymind/status/1173672982453071878

It is a strange import for sure. Maybe we should put a comment above it

daviddq:
Fixded. Sorry for that.

Nice feature BTW

blurymind:
Thank you :) I will do some more testing and get this merged soon

electron/main.js
```diff
@@ -14,7 +13,7 @@ const BrowserWindow = electron.BrowserWindow;
 // be closed automatically when the JavaScript object is garbage collected.
 let mainWindow;
 let yarnRunnerWindow;
-let yarnVersion = version;
+let yarnVersion = app.getVersion();
```
blurymind, line 16: this is better :+1:

daviddq: It was a `var` and I respected the `var` 😆 

Of course we can. I'll change that this afternoon.

blurymind: Theres a lot of this var business still going around Yarn and it its there because in the beginning it didn't use webpack and all dependencies were statically imported from a folder.
Last year I refactored it to get stuff from npm and use webpack, so now we should be allowed to use const and let and the modern goodness :)

src/js/index.js
```diff
@@ -25,6 +25,7 @@ import "lightweight-emoji-picker/dist/picker.js";
 import spoken from "spoken";
 
 import { App } from "./classes/app.js";
+import { version } from "../../package.json";
 
-window.app = new App("Yarn", "0.4.1");
+window.app = new App("Yarn", version);
```
blurymind, line 30: hmm, will this still work in the deployed app? It would suggest that package.json is somehow getting bundled by webpack :thinking: 
Wouldit make more sense to have this version data be in some json file in the public folder and have the new fancy script update is there

daviddq: I don't know what happens with package.json when webpack bundles the webapp. Electron was stripping out the file, that's why I ended using app.getVersion()

If package.json is not available at runtime after bundling, I can:

a) Create a .json in public folder as you mention
b) Replace the value directly on the `window.app = new App("Yarn", "0.4.1");` in index.js

I think a) is more elegant and safe.

electron/main.js
daviddq: Fixed.

src/js/index.js
blurymind: go with a) but dont create a folder. Just a file that holds metadata that changes, such as the version number and if you want to go even fancier - a build stamp (build date or commit?)

daviddq: I've implemented option a)

At build time the version in ./package.json will be copied to ./electron/package.json and ./src/public/version.json

There's still the need for someone to decide to update the version. In the near future the CI system will do that. For now, whenever there are changes on _master_ remember to do:

`npm udpate-version`

That will update the version in all 3 files. Then commit and push the files with the new version.

blurymind: very nice :) This will really come in handy for the continuous integration. I will give it a good test today and get it merged asap, so you can use it in your other PR if you want to :+1:

daviddq: > ...so you can use it in your other PR if you want to 👍

Yes please.

blurymind:
Looks good to me :+1:  Thank you for working on this! Makes it very easy and convenient to bump that number. :) 

merged :tada:

src/js/index.js
blurymind: After  some investigation I discovered that this is what broke hot reloading in webpack. If you edit index js and revert the changes you made to it in this PR, hot reloading in the dev server starts working again

blurymind: actually its not this, there must be something else. Its kind of flaky- sometimes work and sometimes doesn't, which makes it hard to identify.

Another suspect I have is the web-watchers are not doing their job. Some people complained about at the webpack forum. It's a linux specific issue, so we are both getting it

blurymind: Ok I merged a fix that should hopefully fix the hot reload problem. Can you confirm if it does :)

daviddq: Unfortunately the problem seems to persist. It is something that currently prevents me from working on Chrome in Windows because even if the content of the web app changes, the browser continues to show me the cache of hours ago, even if I delete it, it reappears.

On Mac OSX I don't have such a problem.

blurymind: perhaps a regression caused by recent chrome update? :thinking:

PR #141, opened 2020-04-05
Fixes issue #127 Navigating back creates new nodes

daviddq:
Fixes issue #127 Navigating back creates new nodes

### DIFF
--- a/src/js/classes/app.js
+++ b/src/js/classes/app.js
@@ -1311,12 +1311,12 @@ export var App = function(name, version) {
   };
 
   this.openLastEditedNode = function() {
-    if (self.nodeVisitHistory.length < 2) {
+    if (self.nodeVisitHistory.length === 0) {
       self.saveNode();
-      return;
     } else {
-      self.nodeVisitHistory.pop();
-      self.openNodeByTitle(self.nodeVisitHistory.pop());
+      const title = self.nodeVisitHistory.pop()
+      self.propagateUpdateFromNode(self.editing());
+      self.openNodeByTitle(title);
     }
   };
 
@@ -1616,7 +1616,7 @@ export var App = function(name, version) {
       const editorTitleElement = $('#editorTitle')[0];
 
       // Ensure the title is unique
-      const title = self.getUniqueTitle(editorTitleElement.value.trim());
+      const title = self.getFutureEditedNodeTitle();
 
       // Update the title in the UI
       editorTitleElement.value = title;
@@ -1624,9 +1624,6 @@ export var App = function(name, version) {
 
       // Remove leading and trailing spaces from the body links
       self.editing().body(self.trimBodyLinks(self.editing().body().trim()));
-
-      // Update NodeVisitHistory just in case the node's title changed
-      self.updateNodeVisitHistory(self.editing());
 
       self.makeNewNodesFromLinks();
       self.propagateUpdateFromNode(self.editing());
@@ -1645,18 +1642,6 @@ export var App = function(name, version) {
       self.config.autocompleteWordsEnabled = autoCompleteWordsButton.checked;
 
       setTimeout(self.updateSearch, 600);
-    }
-  };
-
-  this.updateNodeVisitHistory = function(node) {
-    const currentTitle = node.title();
-
-    if (node.oldTitle !== currentTitle) {
-      self.nodeVisitHistory.forEach(function(_, index, theArray) {
-        if (theArray[index] === node.oldTitle) {
-          theArray[index] = currentTitle;
-        }
-      });
     }
   };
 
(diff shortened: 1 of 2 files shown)


blurymind:
Thank you! And sorry for the hassle. I should have applied that formatting long ago

src/js/classes/app.js
```diff
@@ -1655,6 +1648,18 @@ export var App = function(name, version) {
     }
   };
 
+  this.updateNodeVisitHistory = function(node) {
```
blurymind, line 1651: I wonder if instead of doing this, we can simply not push a title that has been changed when we navigate to another node.

So on line 1220, can we do
```
    const editorTitleElement = $('#editorTitle')[0];
    const title = self.getUniqueTitle(editorTitleElement.value.trim());
    if (!self.nodeVisitHistory.includes(title)) {
      self.nodeVisitHistory.push(node.title());
    }
```
And thus fix the bug without adding another update function to keep track of :)

daviddq: I don't know if I fully understand your proposal. For what I understand, we would move `self.nodeVisitHistory.push(title);` from `editNode()` function to `saveNode()` function, right? But I don't think that solves the problem. For this to work you need to remove the node being edited from the `nodeVisitHistory` stack on `editNode()`... you win nothing since you end up iterating the `nodeVisitHistory` vector and the logic gets more difficult to understand since it's distributed on several functions.

Maybe I'm missing something?

blurymind: No, my proposal is not to move anything. It is to change the if case which when true adds a new item to the array. The problem is that when you alter the title, the old title is added. That causes a title that no longer exists to be there and trigger a creation. To fix the bug, instead of comparing the old title, we compare the new title in that field and only add the new title if it doesn't already exist in the history.

Your fix adds another function that iterates through all nodes, I thought no we can fix this by simply using the correct title there instead.

daviddq: Yes please, push your proposal since I'm not able to visualize the whole picture.

BTW, my proposal does not iterate all nodes. It only iterates the NodesVisitHistory. But if your proposal avoids the problem with no edge cases I prefer it to having to fix the NodesVisitHistory stack.

blurymind: Ok, we can revert the change if we find problems with it anyways :)

blurymind: The current fix also doesnt completely fix it. If you go to a node via the menu, rename the node then go back twice, a node with the old name still gets created

daviddq: Good catch. Please, push your change. I'll take a look tomorrow.

blurymind:
I pushed a fix for this that avoids iterating, also fixes the case where you visit a node via the context menu, rename the visited node and then click on the button to go to the previous node. That was still creating a new node.

Can you check if you find any problems with this? :)

blurymind:
apologies, I just noticed this morning that I forgot to remove a line last night. Should be fixed now :)

daviddq:
I pulled early this morning and saw that it didn't work but I wanted to research on the change this afternoon before coming back to you. Thanks for the fix. I'll try it after lunch.

daviddq:
I tested your solution. It's behaves very different than mine:

Yours:
no-history

Mine:
history

So, your proposal avoids the problem not allowing to navigate back to renamed nodes while mine allows navigating them.

Here you need to make a decission on what the behaviour should be.

blurymind:
It's not supposed to work the way you think. You can even tell by looking
at my code. The back traversal is only used to go back to a node when you
have navigated away from that by using the context menu. See this gif
https://twitter.com/blurymind/status/1072942754039283712?s=09
The idea here is that you stay in the node editor when you write stories

blurymind:
This or is to fix a bug, not to change the design

daviddq:
Yes... I absolutely misunderstood the purpose of the button. I didn't even know there was a context menu 😅 

Go on with your change then and let's close de PR and issue.

BTW, there are no design changes on the PR. What you see on the gifs is my own fork where I've started implementing themes support.

blurymind:
The new theme looks pretty sweet :heart:  Thank you for working on this.
Also for spotting and fixing this bug. Sorry I didn't explain this feature from the start :)

blurymind:
Merged now :tada:

daviddq:
Would it be easy to turn links in the node body into real HTML links that leads you to the linked node? That way instead of:

1) select the link
2) open the context menu
3) click on the "Edit Node: whatever"

The writer could just ctrl+click the link

blurymind:
lets keep the command in the context menu, but also add the ability to mid-click or ctrl+click on it to go the node.

There are a couple of ways we can make these links nicer too
1. turn the mouse cursor into a hand when hovering over a link to better indicate its clickable + add a tooltip with a hint about crtl+clicking
2. subtly indicate to the user when the link doesnt exists as a node and yarn will create a node- maybe make the link use italic text style or something.

These are just small usability touches of course :) they crossed my mind but never had the time to actually do them

daviddq:
The question was "Would it be easy...?" 😆 . I don't know that editor you use and I don't know how much support I can expect from it to achieve this.

blurymind:
ah right, to do this you need to check out yarn-mode.js where ace editor is set to do the highlighting of text
https://github.com/YarnSpinnerTool/YarnEditor/blob/master/src/public/mode-yarn.js
https://github.com/YarnSpinnerTool/YarnEditor/blob/master/src/public/theme-yarn.js

if you know how to customize ace-editor, it probably wont be hard to you. mode-yarn has access to the methods in the app class, so it will be easy to check if a node in a link exists or not.
Adding some cursor: hand to the link class in theme-yarn.js is probably super easy 

Hope that helps :)

PR #145, opened 2020-04-07
Save the node being edited before editing a different node

daviddq:
This PR prevents this issue:

savenode


package.json
```diff
@@ -16,7 +16,7 @@
     "build": "npm run copy-version && webpack -p --progress --mode production --config webpack.config.js",
     "build-dev": "npm run copy-version && cross-env NODE_ENV=dev webpack -p --progress --config webpack.config.js",
     "copy-version": "node ./scripts/copy-version.js",
-    "copy-web-app": "cd dist && copyfiles *.html ../electron/app && copyfiles **/* ../electron/app && copyfiles **/*/** ../electron/app && copyfiles **/*/**/* ../electron/app",    
+    "copy-web-app": "cd dist && copyfiles *.html ../electron/app && copyfiles **/* ../electron/app && copyfiles **/*/** ../electron/app && copyfiles **/*/**/* ../electron/app",
```
blurymind, line 19: Prettier seems to be formatting files you havent edited. Can you set it to format only on save?

src/js/classes/app.js
```diff
@@ -1142,6 +1142,11 @@ export var App = function(name, version) {
       return;
     }
 
+    // Make sure we save the currently node being edited before editing a new
+    // one using the context menu
+    if ( self.editing() &&  self.editing()!== node )
```
blurymind, line 1147: nit: Some strange spacing there.. is prettier acting strange for you? They recently updated it and now when you format on some editors- it adds a space before and after () of functions

src/js/classes/app.js
```diff
@@ -1638,6 +1639,14 @@ export var App = function(name, version) {
       self.config.autocompleteWordsEnabled = autoCompleteWordsButton.checked;
 
       setTimeout(self.updateSearch, 600);
+
+      // Close editor. SaveNode and CloseEditor should be different functions
```
blurymind, line 1643: you can split it into its own function if you want to reuse it, however lets keep the behaviour of closing it when saving by default

src/js/classes/app.js
daviddq: It is. I can't achieve the same results you achieve with the settings you proposed. I ended up disabling it so it doesn't reformat the whole documents 😞  I'm fixing this.

package.json
daviddq: Yes. I noticed it. I think this comes from the merge of the branches master->save-fix. Anyway, these spaces were unnecessary. Not a big deal.

src/js/classes/app.js
daviddq: Done!

src/js/classes/app.js
daviddq: I think in the future it will be interesting to be able to both

- Save the node, without closing the editor
- Close the editor without saving the node

That's why I wrote that note for the future. Nothing that needs to be addressed now.

SaveNode() still closes the editor by default.

src/js/classes/app.js
```diff
@@ -1142,6 +1142,11 @@ export var App = function(name, version) {
       return;
     }
 
+    // Make sure we save the currently node being edited before editing a new
+    // one using the context menu
+    if (self.editing() && self.editing()!== node)
```
blurymind, line 1147: missed a space before !==

src/js/classes/app.js
```diff
@@ -1607,7 +1612,7 @@ export var App = function(name, version) {
     });
   };
 
-  this.saveNode = function() {
+  this.saveNode = function(closeEditor=true) {
```
blurymind, line 1615: nit:spaces missing before and after =

src/js/classes/app.js
daviddq: Fixed. Please, don't be too picky with this.

src/js/classes/app.js
daviddq: Fixed.

src/js/classes/app.js
blurymind: ah  no worries, sorry for the pickiness. I will try to work out a format command to do it instead

blurymind:
Thank you for working on this  :+1: 
Merged now :tada:

PR #146, opened 2020-04-07
Update arrows while moving all nodes

daviddq:
Fixes the problem of arrows not updating when dragging all nodes:

drag

NOTE: instead of dragging all nodes node-by-node we should be moving the `.nodes-holder`. The performance improvement is amazing. Unfortunatelly, most of the mouse position calculations, offsets, etc. should be reviewed. The whole coordinate system would need a good implementation to abstract the details of screenToWorld and worldToScreen coordinates conversion.


blurymind:
Looks good tome. Thanks for the fix :)
merged :tada: 

on your note on optimisations, feel free to submit another PR if you want to implement it :)

blurymind:
You are getting warmer with the Twine theme btw. Really liking those rounded corners on the nodes :)

daviddq:
Twine-like theme is just a quick test. I'm checking how customizable is the current HTML using css only.

You mentioned you wanted to add some more options to the menu. Can you share your plans on that?

Regarding the optimization, to achieve that it's probably more convenient to start with a code refactoring to try and isolate code related with the canvas rendering. The whole project is in need of a deep refactoring identifying the different entities and functionalities, splitting code into modules, etc. That's very needed but not urgent, so I will keep it on hold.

PR #149, opened 2020-04-09
Arrows redraw simplified and optimized

daviddq:
I've refactored the code related to arrows redraw. I've created a new class Workspace representing the user's workspace. For now I've only moved arrows redraw code there, but I plan to move there everything related to the workspace (marquee, nodes, arrows handling mainly).

I've optimized the updateArrows() function by:

- Drawing lines and arrows in batches (only 2 beginPath() calls per frame)
- Moving context changes out of the nested loops (I sacrificed the arrows transparency)
- Pre-calculating as much as I could imagine out of the nested loops
- Removing the updateArrowsThrottled. Now we only use timed redraws on animations. If we could get an `onFrame()` event on the .transition() method we could avoid this.

As a result arrows drawing is now fluent and is no longer a bottleneck although we still have some others:

arrows

### DIFF
--- a/src/js/classes/app.js
+++ b/src/js/classes/app.js
@@ -5,14 +5,16 @@ import {
   load_dictionary,
 } from '../libs/spellcheck_ace.js';
 import { Node } from './node';
+import { Workspace } from './workspace';
 import { data } from './data';
 import { yarnRender } from './renderer';
 import { Utils, FILETYPE } from './utils';
 
 export var App = function(name, version) {
-  var self = this;
+  const self = this;
 
-  // self
+  this.workspace = new Workspace(self);
+
   this.instance = this;
   this.data = data;
   this.name = ko.observable(name);
@@ -21,7 +23,6 @@ export var App = function(name, version) {
   this.deleting = ko.observable(null);
   this.nodes = ko.observableArray([]);
   this.cachedScale = 1;
-  this.canvas;
   this.context;
   this.nodeHistory = [];
   this.nodeFuture = [];
@@ -49,19 +50,7 @@ export var App = function(name, version) {
   this.hasTouchScreen = false;
   // this.fs = fs;
 
-  this.UPDATE_ARROWS_THROTTLE_MS = 16;
 
-  this.timer = undefined;
-  this.startUpdatingArrows = () => {
-    self.stopUpdatingArrows();
-    self.timer = setInterval(self.updateArrows, this.UPDATE_ARROWS_THROTTLE_MS);
-  };
-  this.stopUpdatingArrows = () => {
-    if (self.timer) {
-      window.clearInterval(self.timer);
-      self.timer = undefined;
-    }
-  }
 
   // this.parser = new ini.Parser();
   this.configFilePath = null;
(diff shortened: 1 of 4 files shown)


daviddq:
I plan to also refactor everything related to input management to an Input class and put some order there. Do this kind of refactoring make sense to you?

blurymind:
The refactoring is definitely something that yarn needs. Splitting the giant app class into other classes makes a lot of sense to me yes.
Will need to test this for regressions and have a more careful look tomorrow, but it looks good :)

blurymind:
I was wondering whether to try memoizing some of the more expensive functions. There are some libraries on nom for that but I haven't tried it yet

daviddq:
Memoizing does a lot of sense to me. So far I've been just fixing the more evident things using intuition and experience, but eventually we will need to to a proper profiling to push performance to the next level.

Regarding story preview, is it there an issue where I can report the bugs/proposals I find?

blurymind:
> Regarding story preview, is it there an issue where I can report the bugs/proposals I find?

Yeah sure,raise issues when you find them.

Regarding performance on this refactoring pr, there seems to be a notable delay now when you do box selection. It takes yarn several seconds to mark nodes as selected after you drew the box.

Box selection with the stress  file is much faster on master -almost instant.

daviddq:
Yes... I noticed it. I'll take a look as soon as I finish my current changes.

daviddq:
I think it's now fixed. Not a fault of the redrawing code but the crazy messed up input events setup. Thanks for the regression checks 👍

blurymind:
The speed increase is very hard to notice. Slightly faster when moving them around. Panning is still very sluggish on both.

But I like the refactor. Do you want to try to improve performance more or should I merge this? :) Thanks for the work on it so far btw

blurymind:
splitting all the node drawing logic into  to workspace.js will pave the way to some day replacing it with something better that can handle many nodes better

blurymind:
Ok I will merge this as it and further performance improvements will be done in another PR.

Thank you for this. :+1:  merged :tada:

daviddq:
> ...The speed increase is very hard to notice

When testing with the stressTest graph yes, speed is similar (crap). But if you create simpler graphs you can notice how dragging some nodes feels more fluent, less _laggy_. For me it's a lot more pleasant.

> Panning is still very sluggish on both.

Yes, there are additional bottlenecks involved in panning unrelated to the arrows drawing the main one being a misunderstanding on what panning is. Panning is not "moving all nodes on a direction" but moving the viewport on the opposite direction... It's like moving a camera... one doesn't move all the objects on the scene, just the camera.

To fix this, as I mentioned some days ago, we should translate and scale `.nodes-holder`, but right now this causes problems when guessing the mouse position.

Once Workspace class is refactored maybe we can dig deeper on this.

blurymind:
there is now some memoization in the arrow drawing logic, please let me know if it helps out a bit :)

blurymind:
There is something wrong with the interval, it always comes up as undefined

daviddq:
What interval are you talking about?

PR #155, opened 2020-04-13
Improve rendering tags allowing styles

daviddq:
Avoid duplicated tags. Generate tag "span"s the knockout way. Apply same style to title and tags for visual candy:

image

image

In the future each tag should have its own style, but that's not yet supported.


blurymind:
very nice refactor. There is a problem with tags disappearing

if you enter 
`er dd erf `
in the tags field, erf will disappear

if you enter 
`er dd er`
er shows up twice


I also dont like deleting tags entered twice in the field - altering the data in the file not to allow having a tag more than once.

I am ok if you show the tag once in the node, but when you open the node, much rather allow the user to enter it more times and save their file.

One improvement that would be amazing is to have a way of counting how many times a tag has been used when you add it while editing the node, but that would require a much bigger refactoring. The tag field as it is atm is missing an autoformatting that puts a circle around each entered tag - like twine has.

daviddq:
My first implementation was to show it once without modifying the data. Then I thought: why to store it duplicated? I'm still thinking of a use case where I want to tag something twice with the same tag. But I'm OK with your reluctance. This is something we can face in the future if you are more convinced.

blurymind:
> My first implementation was to show it once without modifying the data. Then I thought: why to store it duplicated? I'm still thinking of a use case where I want to tag something twice with the same tag. But I'm OK with your reluctance. This is something we can face in the future if you are more convinced.

I am ok with hinting to the user that the tag has been used twice without alarming them (when in the node editor they get circled like twine's), but definitely not ok with limiting them to what they enter there and changing it for them without any notice.

Imagine someone wants to use the tags in a sequence of tags that some might repeat in that sequence - so in their game engine they can trigger some weird cut scene/gameplay/story logic. You are effectively clipping the wings of that and making yarn's export data less flexible

daviddq:
It's changed. With the new implementation I can't reproduce the "er dd erf" issues you mention.

daviddq:
Your example has clarified your point. I have underestimated the power and usability of tags. Now I'm 100% sure we don't have to remove duplicated ones... maybe not even on the display. Tell me if you wan't me to remove that as well.

src/js/classes/node.js
```diff
@@ -42,8 +42,8 @@ export var Node = function(options = {}) {
 
   // clipped values for display
   this.clippedTags = ko.computed(function() {
-    const tags = self.tags().split(' ');
-    return ((tags.length===0) || (tags.length===1 && !tags[0])) ? null : tags;
+    const tags = self.tags().split(' ').filter(item => item);
+    return [...(new Set(tags))]
```
blurymind, line 46: I am ok with removing duplicated tags here though, but without mutating yarn's data. We are just doing it for the individual node display, so  I think you can do

` return [...(new Set(tags..filter((a, b) => tags.indexOf(a) === b)))]`

havent tested

blurymind:
In your screenshot, tags are wonderfully coloured and styled, but when I test the PR's branch - their style is completely missing

src/js/classes/node.js
blurymind: actually not sure now if we should do. Lets leave it for now

daviddq:
> In your screenshot, tags are wonderfully coloured and styled, but when I test the PR's branch - their colour is completely missing - its all grey
> 
> tagStyles

That's weird. It works for me on this branch with no change. Could you please clean your browser's local storage just in case you suffer the same problem than me?

blurymind:
> > In your screenshot, tags are wonderfully coloured and styled, but when I test the PR's branch - their colour is completely missing - its all grey
> 
> That's weird. It works for me on this branch with no change (MacOS 10.15.4, Chrome 80.0.3987.163). Could you please clean your browser's local storage just in case you suffer the same problem than me?

I am suspecting chrome's recent update might have something to do with it.
I will clear my cache and test it again today :)

daviddq:
> I am suspecting chrome's recent update might have something to do with it.
> I will clear my cache and test it again today :)
> 
> Btw the dev server is handled by webpack, so it could be something to do with the webpack settings

I would love to clarify this and find a way to avoid it since it hinders my development cycle a lot. Maybe it has something to do with OfflinePlugin on webpack config, but I'm not familiar with Webpack at all.

blurymind:
> > I am suspecting chrome's recent update might have something to do with it.
> > I will clear my cache and test it again today :)
> > Btw the dev server is handled by webpack, so it could be something to do with the webpack settings
> 
> I would love to clarify this and find a way to avoid it since it hinders my development cycle a lot. Maybe it has something to do with OfflinePlugin on webpack config, but I'm not familiar with Webpack at all.

I will try to fix this as its a recent regression.

blurymind:
I tried clearing the local storage and still get the same result.

Btw not changed from the other PR you made are conflicting with this one

blurymind:
This is strange but the hot reloading work fine on master  and doesn't on this PR for some reason

src/scss/style.css
```diff
@@ -510,8 +510,6 @@ a:hover {
   padding: 4px 8px 4px 8px;
   margin: 4px 4px 0px 0px;
   font-size: 0.8em;
-  background: #e4dbcb;
-  color: #978e7e;
```
blurymind, line 514: have you moved this to somewhere else? If so, I dont see it in this pr

src/index.html
```diff
@@ -120,7 +120,9 @@
 					<div class="node" data-bind="nodeBind: true, css: { inactive: !node.active()}" >
 						<div class="title" data-bind="text: node.title, css: 'title ' + node.titleStyles[node.colorID()]"></div>
 						<div class="body" data-bind="html: node.clippedBody"></div>
-						<div class="tags" data-bind="html: node.clippedTags"></div>
+						<div class="tags" data-bind="visible:node.clippedTags, foreach: { data: node.clippedTags, as: 'tag' }">
+							<span data-bind="text: tag, css: node.titleStyles[node.colorID()]"></span>
```
blurymind, line 124: can we not do a for each inside the html file? I prefer to have this happen in some of the js files

src/scss/style.css
daviddq: Now tags' color/background are dynamically styled by .title-style-X {} styles

blurymind: I suspect its not working because you moved some of their rendering js code to the html file

src/index.html
daviddq: This is the proper way of unrolling tags if you want them to be bound by Knockout to their style and be able to switch it dynamically.

src/scss/style.css
daviddq: It still works on my side out-of-the-box, just checking out the branch and running the webapp.

Regarding index.html, I'm changing it for "settings dialog" and for "help dialog" as well and it's refreshing OKish.

daviddq:
Did you know about this branch by @ashildebrandt:

image

It already has theme-based dark mode, snap to grid, unified search, menu, quick actions icons, etc.

daviddq:
Oh man! Now I see the "problem": tags have the same style than the title. You need to change the title color to change the tags color:

Captura de pantalla 2020-04-14 a las 20 01 21

It is a purely aesthetic change. As I mentioned on first comment:

> In the future each tag should have its own style, but that's not yet supported.

But now there's no support for coloring each tag with a different color.

blurymind:
> Did you know about this branch by @ashildebrandt ?
> 
> image
> 
> image
> 
> It already has theme-based dark mode, snap to grid, unified search, menu, quick actions icons, etc.
> 
> Is this kind of menu something you consider? I would like to give some consistency to the menu and general layout. That's something that goes beyond .css tweaking since HTML markup should be arranged.

I did not :) it would be nice if we can get those features from his branch into ours, but without the total restyling he has made to yarn

src/scss/style.css
blurymind: Ok :) just hope we dont get webpack problems

src/index.html
blurymind: right, to be honest I don't know knockout very well.My understanding of that library is surface level :)

daviddq:
> Oh man! Now I see the "problem": tags have the same style than the title. You need to change the title color to change the tags color:
> 
> Captura de pantalla 2020-04-14 a las 20 01 21
> 
> It is a purely aesthetic change. As I mentioned on the first comment:
> 
> > In the future each tag should have its own style, but that's not yet supported.
> 
> But so far there's no support on the data model for coloring each tag with a different color.

Did you try again? Does it work?

blurymind:
> > Oh man! Now I see the "problem": tags have the same style than the title. You need to change the title color to change the tags color:
> > Captura de pantalla 2020-04-14 a las 20 01 21
> > It is a purely aesthetic change. As I mentioned on the first comment:
> > > In the future each tag should have its own style, but that's not yet supported.
> > 
> > 
> > But so far there's no support on the data model for coloring each tag with a different color.
> 
> Did you try again? Does it work?

As I see it the pr has two problems:
- its not serving its own title and screenshot in terms of delivering this to the end user - not adding styling of tags. If tags don't have that, you would have to add it :)
- It has merge conflicts

Why would it wield different results if I try it again? There have been no new commits since I tried it before :)

daviddq:
Did you try to change the node's title color?

blurymind:
> Did you try to change the node's title color as I suggested? styled-tags

Ah I see, so this is actually intended by the PR? If so, I am ok with that, but the default tag colour doesnt look very good, its sort of impossible to see the box. I see that as a regression from what we have on master by default
Screenshot_2020-04-15_15-57-53

daviddq:
Fixed:

image

Also merge conflicts are fixed. Again on my side it detects the whole app.js and node.js fully changed when only 5 to 10 lines were changed :(

src/js/classes/app.js
```diff
@@ -1,2106 +1,2106 @@
...
-
-    for (; ; ++counter) {
-      var newTitle = baseTitle + '_' + counter;
-      if (!currentlyUsedTitles.includes(newTitle)) {
-        return newTitle;
-      }
-    }
-  };
-};
+import {
```
blurymind, line 1: we really need to find a way to fix your code editor from doing this. It will only lead to more merge conflicts

blurymind:
> Fixed:
> 
> image
> 
> Also merge conflicts are fixed. Again on my side it detects the whole app.js and node.js files fully changed when only 5 to 10 lines were changed :(

Thank you for making this look nice again :)

Only one last thing - can you please fix the app.js file from being completely diffed by this pr? It's gonna make merge conflicts for both you and me in the next one, so this is starting to become inconvenient

src/js/classes/app.js
daviddq: I think it wasn't me. The problem came from integrating your changes in _master_ to my branch _style-tags_

I use VSCode with no prettifier. The only thing it does on save is remove trailing spaces.

blurymind: I haven't done any changes to this branch, you can see no commits from me. I don't understand how I could be doing this. :)
I also use vscode but now disabled prettier so as to avoid stepping on your toes due to formatting.

does every line of app.js has a trailing space :thinking: 
Is that the thing causing this

daviddq: I wrote that the problem showed up when I integrated "_your changes in master_" into my branch. You have indeed made changes to _master_:

image

That doesn't mean that you made a mistake or something. I'm simply pointing when I think that the problem showed up.  Why? No clue because reviewing changes to app.js on the history I can't see such a change on all lines. Maybe there's some literature about this issue on the Internet. I'll take a look.

daviddq: OK. I think I found it:

you used Windows style line endings for commiting app.js:

Captura de pantalla 2020-04-16 a las 11 40 46

while I used Unix style:

Captura de pantalla 2020-04-16 a las 11 41 37

I can configure my repo settings to use Windows style line endings if that's what you use, although the widely accepted standard is Unix line endings.

Please, confirm.

blurymind: thats a good catch! Let me check again, it's strange as I am using manjaro linux with vanilla settings. Why is vscode enforcing those settings and who else has edited that file recently? :nerd_face: 
How do I check the settings? Is that in vscode?

blurymind: Ok I just notice now that yarn's line endings on files is a bit of a mess atm. We need to unify them all in another pr. CRLF or LF - whichever you prefer

daviddq: On one side there's Git converting line endings:

a) when pulling a file from the repot to your machine
b) when pushing a file from your machine to the repo

And then there's VSCode converting line endings whenever you:

c) open a file
d) save a file

Regarding Git, you can see/change your config with:

`git config core.autocrlf`

on your terminal.

Regarding VSCode, I'm looking for a way to change the setting only for a project.

daviddq:
For now I converted app.js and node.js to CRLF line endings so it's properly _diffed_ and we can close this PR.

blurymind:
> For now I converted app.js and node.js to CRLF line endings so it's properly _diffed_ and we can close this PR.

Thats ok, we can fix it in all files in another commit. Thank you for doing that. Do you prefer CRLF or LF? 

Also I left some notes about the style

daviddq:
I certainly prefer LF since it's a _de facto standard_.

> Also I left some notes about the style

Where?

daviddq:
Strangely I had to convert app.js to CRLF and node.js to LF for GitHub's diff to work OK

blurymind:
> I certainly prefer LF since it's a _de facto standard_.
> 
> > Also I left some notes about the style
> 
> Where?

https://github.com/YarnSpinnerTool/YarnEditor/pull/155#issuecomment-614504578

daviddq:
Oh! OK. I didn't notice your styling comments before.

The thing is that node title and tags use now the same dynamic style. This is the closes I can get to the previous style since changing one changes the other:

image

Is it OK or should I cancel the PR?

blurymind:
> Oh! OK. I didn't notice your styling comments before.
> 
> The thing is that node title and tags use now the same dynamic style. This is the closes I can get to the previous style since changing one changes the other:
> 
> image
> 
> Is it OK or should I cancel the PR?

Can you make the title and tag background slightly lighter :)

daviddq:
As you mention having different text color for title and tags is no longer possible. If that's an impediment maybe it's better to close the PR.

blurymind:
Why do you need to enforce it in js to have them share the style? With little change to the pr you can add a class on the node itself to each different coloured style and via that class optionally apply a colour to the tag and the title independently.  That way we keep the old theme exactly the same as before and your new theme can have dynamic tag colours.

Alternatively you can just have tagstyle1 tagstyle2 and so on, a separate class than the title. That way this is optional in the css and not enforced on all themes.

daviddq:
Yes... binding the tags style to the title style was probably not the best idea.

However, I don't see providing each different tag a different class (and it to be consistent across all nodes) a little change. We should keep track of all the tags in the story and I haven't seen such a feature so far.

blurymind:
I bet there is a simple way to bind a new class to tags the same way you do
with title. Let me see if I can do something to this pr or it's a huge
refactoring warranting closing it for something else

daviddq:
The problem is not binding a new class. The "problem" is that if I use "tag1" on node "Start" and "tag1" on node "Node_1" both share the same style. Right now there's no registry of all tags available on the story.

daviddq:
I'll implement a "tags" repository. That way we can know:

a) what tags are currently used on the story
b) the number of times a tag is used

You like b), right? That way we will be able to show the dot with the number of appearances next to the tag on a future PR.

daviddq:
It's done. Now tags can remain the same if they are not styled:

image

Or they can be styled independently from the title:

image

src/js/classes/app.js
```diff
@@ -34,6 +34,7 @@ export var App = function(name, version) {
   this.editing = ko.observable(null);
   this.deleting = ko.observable(null);
   this.nodes = ko.observableArray([]);
+  this.tags = ko.observableArray([]);
```
blurymind, line 37: nice work :+1:  the data model is handling tags way better

src/js/classes/app.js
```diff
@@ -505,7 +506,10 @@ export var App = function(name, version) {
         if (!self.previewStory.finished)
           switch (e.key) {
             case 'z': {
-              self.advanceStoryPlayMode();
+              self.previewStory.changeTextScrollSpeed(200);
```
blurymind, line 509: doesnt seem related to this PR but I see you also want to fix some bugs with the previewer :) 
That means I have to test it again now

blurymind:
Ok this looks great now! :+1: Thank you for working on it and also for following up the feedback so quickly. :)

Fantastic work on cleaning up the code and putting the especially messy parts in order

blurymind:
Merged to master :tada:

src/js/classes/app.js
blurymind: This opens the door for some exciting features, such as tag autocompletion as you type them(like twine's) and even storing the tags array in the exported json file - so game engines can do some interesting things with that data without having to laboriously collect it from all nodes

daviddq: Correct! Tags are now data model's first class citizens 🤣

src/js/classes/app.js
daviddq: I am not aware of having made that change. Didn't it come to my branch when integrating a change in master? 🤔

daviddq:
For the next PR I have two candidates:

- Improve file loading time
- Refactor of the input related stuff to Input class

Which one do you prefer first? 😄

src/js/classes/app.js
blurymind: I fixed the 
npm run format 
command, so we can have consistent formatting on files

src/js/classes/app.js
blurymind: that is a bit strange :) yeah I think you are right.
I recently edited that bit of logic to fix some bugs with the previewer,  has this somehow undone my changes or was it wrongly indicating them in the dif?

blurymind:
They both sound like quality material. The loading time will be beneficial to end users more, so lets go with that one :)

Why not just put them both up at once

src/js/classes/app.js
blurymind: arrgh, this is annoying, it has. I need to undo this bit now

blurymind: ok I cherry picked and reverted it in another commit

PR #159, opened 2020-04-20
Improve load time

daviddq:
Reduced the time to load a graph from file.

app.js and data.js fully changed do to `npm run format`


blurymind:
interesting, its still thinking the files are different.. I just ran format on master and merged the changes inflicted by it :thinking:  I think eslint should have unified the line endings to be the same

blurymind:
can you do
`npm run format`
on this branch and merge all the changes you get please?

daviddq:
Done!

blurymind:
Beautiful! Thank you very much :+1: now we have proper linting and formatting working again and no more weirdness in difs.
Let me know if you want any of the eslintrc settings changed :) I am happy with any suggestions

daviddq:
So before every push I must run "npm run format" and we will be safe, right?

src/js/classes/data.js
```diff
@@ -84,39 +83,37 @@ export var data = {
...
-    if (filename.toLowerCase().indexOf('.json') > -1) return FILETYPE.JSON;
-    else if (filename.toLowerCase().indexOf('.yarn.txt') > -1)
-      return FILETYPE.YARN;
-    else if (filename.toLowerCase().indexOf('.yarn') > -1) return FILETYPE.YARN;
-    else if (filename.toLowerCase().indexOf('.xml') > -1) return FILETYPE.XML;
-    else if (filename.toLowerCase().indexOf('.txt') > -1) return FILETYPE.TWEE;
-    else if (filename.toLowerCase().indexOf('.tw2') > -1) return FILETYPE.TWEE2;
-    else if (filename.toLowerCase().indexOf('.twee') > -1)
-      return FILETYPE.TWEE2;
+    var lowerFileName = filename.toLowerCase();
```
blurymind, line 86: :+1:  nice, thanks for cleaning this up

blurymind: actually can you make this a const, we are not changing  so no need to be a var

daviddq: Sure thing! Done!

src/js/classes/data.js
```diff
@@ -84,39 +83,37 @@ export var data = {
...
-      return FILETYPE.YARN;
-    else if (filename.toLowerCase().indexOf('.yarn') > -1) return FILETYPE.YARN;
-    else if (filename.toLowerCase().indexOf('.xml') > -1) return FILETYPE.XML;
-    else if (filename.toLowerCase().indexOf('.txt') > -1) return FILETYPE.TWEE;
-    else if (filename.toLowerCase().indexOf('.tw2') > -1) return FILETYPE.TWEE2;
-    else if (filename.toLowerCase().indexOf('.twee') > -1)
-      return FILETYPE.TWEE2;
+    const lowerFileName = filename.toLowerCase();
+
+    if (lowerFileName.includes('.json')) return FILETYPE.JSON;
```
blurymind, line 88: maybe we should change this to endsWith :thinking:

src/js/classes/app.js
```diff
@@ -92,6 +92,13 @@ export var App = function(name, version) {
     window.addEventListener('touchstart', function() {
       self.hasTouchScreen = true;
     });
+    window.addEventListener('yarnLoadedData', e => {
```
blurymind, line 95: crazy stuff. Was it actually updating arrows while the file was loading :facepalm:

src/js/classes/data.js
```diff
@@ -225,37 +222,25 @@ export var data = {
...
-    if (numAvg > 0) {
-      app.workspace.warpToXY(avgX / numAvg, avgY / numAvg);
+    app.nodes.extend({ rateLimit: { method: 'notifyWhenChangesStop', timeout: 250 } });
+    {
+      if (clearNodes)
+        app.nodes.removeAll();
+
+      objects.forEach( (object, i) => {
+        let node = new Node({
+          title: object.title,
```
blurymind, line 232: interesting, why was it checking if they were undefined before? Hope theres no corner case where the json file is missing a key, causing the app to crash here. Maybe we should have a protection for that

src/js/classes/data.js
daviddq: Done!

src/js/classes/data.js
```diff
@@ -225,37 +222,25 @@ export var data = {
...
+          body: object.body,
+          tags: object.tags,
+          colorID: object.colorID,
+          x: object.position.x,
+          y: object.position.y,
```
blurymind, line 237: why was it parsing the int before?
Can you check if loading xml and other supported formats still works fine?

src/js/classes/app.js
daviddq: Not exactly. Each time something changes on Knockout's data model it triggers some bindings. This bindings trigger computed values, HTML DOM changes (super expensive) and arrows drawing. All together makes it a lot of useless overhead **each time a single node is added to the data model**.

blurymind: this is a nice catch :+1:  I will test with the stress file again :)

src/js/classes/data.js
```diff
@@ -84,39 +83,37 @@ export var data = {
...
-    else if (filename.toLowerCase().indexOf('.yarn') > -1) return FILETYPE.YARN;
-    else if (filename.toLowerCase().indexOf('.xml') > -1) return FILETYPE.XML;
-    else if (filename.toLowerCase().indexOf('.txt') > -1) return FILETYPE.TWEE;
-    else if (filename.toLowerCase().indexOf('.tw2') > -1) return FILETYPE.TWEE2;
-    else if (filename.toLowerCase().indexOf('.twee') > -1)
-      return FILETYPE.TWEE2;
+    const lowerFileName = filename.toLowerCase();
+
+    if (lowerFileName.endsWith('.json')) return FILETYPE.JSON;
+    else if (lowerFileName.endsWith('.yarn.txt')) return FILETYPE.YARN;
```
blurymind, line 89: I guess the else if kind of redundant here, since the return in the previous line will stop it from being checked anyways

blurymind:
well done on this! Testing with the stressTest file, it does seem to load noticeably faster :+1:

daviddq:
Anything missing for the _merge_?

blurymind:
Was just wondering why you removed the checks for undefined values it had
before. :) Why did yarn sanitized the loaded data to begin with?

daviddq:
The Node's constructor does the "undefined" checks already.

blurymind:
Ok, thank you for working on this :+1: 
merged now :tada:  :)

PR #161, opened 2020-04-22
Removed unused code

daviddq:
Removed unused code


src/js/classes/app.js
```diff
@@ -1528,15 +1528,6 @@ export var App = function(name, version) {
     return tagBeforeCursor;
   };
 
-  this.testRunFrom = function(startTestNode) {
-    ipc.send(
-      'testYarnStoryFrom',
```
blurymind, line 1533: you can also get rid of https://github.com/YarnSpinnerTool/YarnEditor/blob/f18777ebe8de5c92b48a31cab687cfffaefe9222/electron/main.js#L93

as well as
https://github.com/YarnSpinnerTool/YarnEditor/blob/f18777ebe8de5c92b48a31cab687cfffaefe9222/electron/main.js#L101

and yarnRunnerWindow references in the samefile

daviddq: I use it as webpage as well, but my narrative designer will use it in Electron.

Code in YarnEditor/electron/* is not _eslinted_. I will keep it unformated. If you want it formated, please, update the "npm run format" script.

daviddq: I think this is unused as well: 

https://github.com/YarnSpinnerTool/YarnEditor/blob/b5a6bdba6c7f51af5218451f7d5382e3a69bb702/src/js/classes/app.js#L1899

Can you confirm?

blurymind: wow, that is one long function. I need to investigate it. Its possible it did something useful in the past, but somehow ended up becoming cut off. Yarn has a few legacy things lying around. If nothing is referencing it in the rest  of the app, thats likely the case

blurymind: Feel free to remove it if you can't think of any use for it. Its completely dead code atm

blurymind:
good work! :+1: 
merged now :tada:

PR #162, opened 2020-04-23
Fix tags

daviddq:
This is a quick one. Ensures that the tags repository is updated before splitting a node's tag. Solves an issue on Windows.


blurymind:
good catch :+1: 
merged now :tada:

PR #163, opened 2020-04-23
Panning

daviddq:
Move .nodes-holder instead of moving all nodes.


src/js/classes/node.js
```diff
@@ -1,14 +1,14 @@
...
-var globalNodeIndex = 0;
+let globalNodeIndex = 0;
 const NodeExpandWidth = 300;
 const NodeExpandHeight = 150;
 const ClipNodeTextLength = 1024;
 const bbcode = require('bbcode');
 
-export var Node = function(options = {}) {
-  var self = this;
+export let Node = function(options = {}) {
```
blurymind, line 10: wondering now if this should be a const :thinking:

blurymind:
Nice work so far!
This seems to have broken the zoom level buttons in the bottom right of the app

blurymind:
Another problem I notice is that if you right click, the browser's context menu will appear, which is not something we want

src/js/classes/workspace.js
```diff
@@ -18,87 +18,147 @@ export const Workspace = function(app) {
...
+  //
+  // Sets the translation on the transform matrix
+  this.setTranslation = function(x, y, time=0) {
+    self.transform.x = x;
+    self.transform.y = y;
+    self.translate(time);
+  };
+
+  // shiftTranslation
+  //
```
blurymind, line 44: Do you leave this gap for future TODOs :)
Interesting habit

src/js/classes/app.js
```diff
@@ -119,7 +112,9 @@ export var App = function(name, version) {
       osName = 'mobile';
     }
 
-    if (osName == 'Windows') self.zoomSpeed = 0.1;
+    // Platform specific settings
+    if (osName == 'Windows') self.workspace.zoomSpeed = 0.1;
+    if (osName === 'mobile') self.workspace.zoom(3);
```
blurymind, line 117: nit: can be else if

blurymind, line 117: Actually, can you remove this line please?
`if (osName === 'mobile') self.workspace.zoom(3);`

It completely breaks yarn on a portrait mode smartphone

daviddq: That line is not new. Are you sure you want to remove it?

blurymind: > That line is not new. Are you sure you want to remove it?

move it back to where it was before

src/js/classes/node.js
daviddq: Yes. I did a quick search and replace var -> let

daviddq:
> Also when you pan, are you still altering all node positions or does this fix that problem? Their coordinate stays the same in the file you save - if it hasnt been moved

There's no need to alter the nodes' positions now, so the diff problem should be mitigated.

src/js/classes/app.js
daviddq: All mentioned bugs fixed. Thanks for helping to test it 😉

blurymind:
Nice work on this. :+1:  and thank you for fixing the regressions

The performance boost is noticeable, the codebase- meaner and cleaner, the dif noise that yarn created before when saving files - gone!

merging this now! :tada:

PR #164, opened 2020-04-24
Minor fixes

daviddq:
Some minor fixes:

- Fixed Enter key detection.

- Fixed typo "sapp" instead of "app"

- Update tags repository only once when a node is edited instead of everytime a node's tags are clipped (several times per node edition).


blurymind:
nice. Thank you for catching these :)

merged :tada:

PR #170, opened 2020-05-01
Added "Blueprint" theme

daviddq:
I added the "Blueprint" theme. Also did some refactor on the .css files.

The .css files were (an are) a little bit of a mess but I didn't want to do a whole refactor since you take so much care of the current appearance. This makes it very difficult to rearrange the styles in a more coherent and useful way.

I moved the "new node" to a button on the icons area. My proposal is to only show this icons on mobile while on desktop we offer that functionality on the main menu:

image

And I fused the two search boxes in just one as we spoke weeks ago.

Please, take a look and report the bugs you find. Thanks.


daviddq:
I'm testing the PR on Windows and it's completely rotten 🤔 

I'll check it and keep you updated.

daviddq:
The issue is caused by this line being added to the index.html:

image

Do you know what is it?

daviddq:
False alarm. The issue was caused by a Chrome extension that injects css overrides into the webapp. It is the extension I used to test the themes before I could change them on the settings 🤦‍♂️

blurymind:
Thanks for this PR.  I will have a look :) Code looks alright

blurymind:
I really like this. Moving the add new node button there is brilliant. Also the way you unified the menus.

I have only one minor gripe - chrome's annoying autocompleter pops up covering our newly moved nodes menu. I wonder if it would be worth making it appear only if the user has entered more than 0 characters in the search field. Tried googling it and couldnt find a solid solution to get rid of that. Maybe we can merge this and try to fix it at a later stage

blurymind:
Ok I found some issues actually, on a mobile screen, the layout is messed up:
messedUpLayout

compare this to:
yarnMobile

and looks kind of broken

Can you make the nodes list menu appear when the search box is in focus or the user is typing in it? Right now it appears if upon hovering the mouse anywhere around it which is not ideal.

daviddq:
I understand the problem with the "tag" button and "File" menu on mobile. I'll fix it 👍 

But I don't understand the problem with the autocompleter. Please, can you attach a screenshot?

src/index.html
```diff
@@ -1,412 +1,433 @@
...
+</head>
+<body>
+	<!-- Fancy Background -->
+	<div id="app-bg">&nbsp;</div>
+
+	<!-- Container -->
+	<div id="app">
+
+		<!-- search form -->
+		<div class="app-search" onmouseenter="app.ui.openNodeListMenu('open')">
```
blurymind, line 16: can you remove this onmouseenter please. You are already triggering it oninput futhrer down

daviddq:
✔️ Search filters (title, body, tags) now stay on place.
✔️ File menu aligned to the left. Reduced vertical gap.
✔️ Search menu now keeps opened while input element is focused
❓  I'm not sure about what the problem is exactly. Please, provide more info and screenshot.

src/index.html
daviddq: They serv different purposes:

- onmouseenter: triggers the menu to show when mouse hovers over the search input. That allows you to quick jump to a node without the need to press any key.

- oninput: updates the content of the search as the user types into the search input.

blurymind: yes I am aware, but there are two problems
- onmouseenter  does not work on mobile devices at all
- it is a bit unexpected for the menu to appear on hover, it also gets triggered when you hover over the buttons (tag,title,etc).

src/index.html
```diff
@@ -1,412 +1,433 @@
...
+			<div class="menu">
+				<input
+					id="nodeSearchInput"
+					type="text"
+					class="search-field"
+					placeholder="Search"
+					oninput="app.ui.openNodeListMenu('open')">
```
blurymind, line 23: `onclick="app.ui.openNodeListMenu('open')"`

and remove onmouseenter on the parent node

daviddq: Do you mean replacing "oninput" by "onclick" ?

blurymind: no- you can have both on one element you know :)

daviddq: Oh! Yes, yes... makes sense 😅

daviddq: Done !

blurymind:
The tag button gets clipped off now
Here's a comparison
clippedHeader

Btw you can use the online version to compare for regressions
https://yarnspinnertool.github.io/YarnEditor/

daviddq:
I removed the margins to gain some extra space.

What's the minimum width for the search bar? 360px?

blurymind:
> I removed the margins to gain some extra space.
Can you please bring it back for the classic theme? 


> What's the minimum width for the search bar? 360px?

Can we not use minimal width and have this be reactive like before?


Also here is what happens when I test still:
Peek 2020-05-03 20-53

- hovering over the menu, chrome's autcompleter covers the menu
- hovering over the tags/title still triggers the menu

daviddq:
> Can you please bring it back for the classic theme?

Sure thing!

> Can we not use minimal width and have this be reactive like before?

As you wish ! :)

> hovering over the menu, chrome's autcompleter covers the menu (still gets triggered onhover)
> hovering over the tags/title still triggers the menu

That's not happening anymore. Please, refresh webapp cache, etc. I researched on that and it's caused by the `offline-plugin` and the service worker it creates.

blurymind:
There is apparently a way to disable autocompleters via a dom prop
https://stackoverflow.com/questions/12374442/chrome-ignores-autocomplete-off

but chrome for whatever reason completely ignores it

daviddq:
I think I finally understood what the problem is with the search box width. Now it's responsive again. Please, validate it.

daviddq:
> There is apparently a way to disable autocompleters via a dom prop...

Yep... I've been reading about it. It's a little bit tricky. If I could only reproduce the "problem" I could find a way to avoid it.

blurymind:
Its responsive, but it still looks worse when compared to master:

Screenshot_2020-05-03_21-15-07

On your chrome debugger,select the tiny smarphone icon, then in the dropdown select moto g4

blurymind:
I also noticed that on master some of the buttons in the bottom right get hidden when we go narrow - the expectation being that you will likely not use them on a smartphone

daviddq:
I know I can check with the Developer Tools.

That's why I asked about the minimum width for the search bar. Your screenshot is taken at 360px width. If you close it a little bit, version 0.4.5 will look like 0.4.8. So:

✔️ It needs to keep the structure
✔️ It needs to be responsive
⌛ At 360px, search bar has fit on the same row
⌛ At 360px, search margins need to mimic the ones at version 0.4.5

right ?

blurymind:
yes, visually preserve the small padding, both on left and right - your padding is huge on all sides and completely cropped on the right side.

Just open https://yarnspinnertool.github.io/YarnEditor/ in another tab and set both yours and that one to the same debugger preview mode (moto g4) and look for things that break

blurymind:
Hope this helps a bit
annotated

daviddq:
Thanks for the visual feedback. It helped a lot.

I think it's done. Correct me if I'm wrong.

I also set the search input type to "search" to avoid autofill / autocompletion. I can't test it since on my side autocompletion doesn't show up.

blurymind:
thanks :) I will have another look in my lunch break :+1:

blurymind:
This is much better, thank you :) 
The autocomplete is still autocompleting, so you can remove the code that was trying to stop it.

There is just one last problem:
Screenshot_2020-05-04_18-22-14

The file menu button is now pushed down by the node list menu. Ideally the menu should be above it and not push it down. If its hard to fix, we can leave it for some other time. Its not a big issue to be fair, just looks a bit funky


We can try to come up with something for the autocomplete at a later stage

daviddq:
Would this gap be OK? 

image

It's needed to place there the search flags if we want it to be responsive:

image

That's the most I can do with just .css

blurymind:
lets leave it as it is for now. I prefer it over the gap :)

blurymind:
Thank you for working on this! :+1: The new theme looks fantastic btw :)

blurymind:
merged :tada:  
Another excellent PR ! :muscle:

PR #171, opened 2020-05-03
Avoid building the project on "pull request" since the GH_TOKEN is not available

daviddq:


daviddq:
Any news on this? This PR is just to avoid triggering the build process each time I push to a PR since this builds will fail due to permissions:

image

blurymind:
Is there a way to  get the other two to build instead? I think I now have access to the settings, so I might be able to helpyou.

Also If I want to make a release at https://github.com/YarnSpinnerTool/YarnEditor/releases how do I do it using this? Can I? Our last release is from ages ago

daviddq:
Build artifacts are automatically published: https://github.com/YarnSpinnerTool/YarnEditor/actions/runs/95524086

Honestly, I don't know how to fix the failing builds. I provide the GH_TOKEN on my fork and you provide GH_TOKEN on your side. I don't know why is it failing. I read on the forums about it and it seems that's a limitation right now, but I'm not 100% sure.

daviddq:
Regarding releases, I would suggest embracing "semantic versioning":

> Given a version number MAJOR.MINOR.PATCH (i.e. 0.5.3) increment the:
> 
> MAJOR version when you make incompatible API changes,
> MINOR version when you add functionality in a backwards compatible manner, and
> PATCH version when you make backwards compatible bug fixes.
> Additional labels for pre-release and build metadata are available as extensions to the MAJOR.MINOR.PATCH format.

In our case:

MAJOR -> changes that introduce possible incompatibilities on the files (i.e. adding "node width" and "node height" to the file)

MINOR -> new compatible features

PATCH -> compatible bug fixes

Also I would suggest hopping to version 1.0.0 since the tool is production-ready.

blurymind:
@desplesda do you know why the other builds might be failing? I think a big chunk of our userbase in on mac and windows. It would be nice if we could set up some automation for the released page to publish the automatic builds from the master branch

daviddq:
@blurymind Notice that the builds that fail are those related to the PRs. Once you merge the PRs into _master_ the resulting builds are OK. You can use those for the release page.

blurymind:
Ok,I am merging this then :)

Its nice we now have a build process, just wonder if we can somehow make it put the stuff where our users will find it
https://github.com/YarnSpinnerTool/YarnEditor/releases

PR #172, opened 2020-05-05
Fix selection

daviddq:
This PR fixes a bug in the marquee selection:

image

Without the fix, a node gets selected only if the top-left vertex of the node is inside the marquee.


blurymind:
thank you for this fix. I will have a look after work :)

blurymind:
well done! :+1:  merged now :tada:

PR #173, opened 2020-05-05
Avoid building on "pull request"

daviddq:
I forgot to push this change to avoid building on every "PR"


blurymind:
merged now :) 

Can you look into adding a stage to the script that publishes the new builds at 
https://github.com/YarnSpinnerTool/YarnEditor/releases

PR #174, opened 2020-05-05
Fix themes in electron builds

daviddq:


blurymind:
If I remember right, electron's packages.json copies the files to a folder before building it, so you don't have to move or copy them manually.. Need to have another look at that :) it's possible that the bud script needs updating so it copies your new themes folder

blurymind:
Actually I do prefer to have them in the public folder. Keep it as it is, will test and merge asap :)

blurymind:
very strange, running npm install in the electron folder is giving me an error

daviddq:
npm install
npm run build-windows

Both work on my side (Windows 10)

blurymind:
ah nevermind. Its an npm cache bug, using yarn works just fine :)

blurymind:
Works on linux too -the build that is :) nice work. :+1: 

merged now! :tada: 

Can you look into making the script publish these at the releases page. It would be awesome and more people will see the new features you worked on

daviddq:
> Can you look into making the script publish these at the releases page. It would be awesome and more people will see the new features you worked on

I don't have a clue on how the release page works, but yes, it's a nice feature: create a Trello ticket and I will handle it eventually.

Issue #175, opened 2020-05-07
add link menu in node editor no longer works


blurymind:
Some of the recent PRs must have broken the add link menu (top right in the node editor)
@daviddq can you have a look please :)
Clicking on it does nothing

blurymind:
It worked on yarn 0.4.5 and is broken on 0.4.10

blurymind:
fixed by @daviddq  so can close now. Thank you for the quick fix :+1:

PR #176, opened 2020-05-07
Fix add link menu

daviddq:
Fixes issue #175


blurymind:
Wow that's fast!
Thank you, I will have a look on lunch break

blurymind:
Thank you for fixing this and sorry for missing it initially. I replaced a onmouseover with an onclick as it was preventing it from working on touch devices.

Merging this :tada: ta

PR #180, opened 2020-05-10
Removed warning

daviddq:
Removes an annoying warning in renderer.js


blurymind:
good catch, thank you! :+1: 
:tada: merged! :)

PR #186, opened 2020-05-25
Changes needed for VSCode webview extension

TranquilMarmot:
# What is this?

This PR contains changes needed for the VSCode extension to have an embedded webview of the editor.

Checkout out the VSCode extension PR here: https://github.com/YarnSpinnerTool/VSCodeExtension/pull/1

The goal of this PR was to leave the editor intact for current use cases (in-browser app and Electron app) but still allow for it to be embedded in a VSCode webview.

This is my first time working with anything Knockout related, so bear with me! I might have done some silly things.

All in all, it was surprisingly easy to shove this into a VSCode extension. The way that the events and everything are hooked up worked nicely for this use case!

# Changes

## Assets in the `public` folder

For more info on this read the "[Loading local content](https://code.visualstudio.com/api/extension-guides/webview#loading-local-content)" section of the VSCode webview docs.

Basically, for security reasons, the webview only has access to specific assets. These specific assets need to be loaded with a URI that is generated by the webview with a scheme of `vscode-resource:`. As you can imagine, there are a lot of hoops to jump through here.

As far as this PR is concerned, this means a new function in `Utils` called `getPublicPath` that can be used to get the path for an asset that would normally be acessible in the `public` folder. Any assets that _were_ referring to `public` are now piped through this (with the exception of some assets in `index.html` that are find-and-replace'd with the proper URI by the extension when it renders the HTML).

## `localStorage` and settings

Accessing `window.localStorage` _at all_ within a webview will immediately throw an exception and cause the app to stop executing. It will now fake an object with noops for getting/setting values.

All settings from inside the extension are instead handled by VSCode's built-in settings mechanisms; see the other PR for more info.

## The "Start" node

This one was a bit of a doozy; you'll see that I added in `app.js`:

```js
if (!window.editingVsCodeFile) {
  self.newNode().title('Start');
}
```

This is for the use case of opening an existing file in VSCode. If opening a file, the extension will set `window.openingVsCodeFile` to `true` so that the "Start" node will not be created.

This has to be done because the extension listens for the `yarnReady` and then calls out to (among other things) `data.loadData`. At this point, `app` would be trying to call `workspace.updateArrows();` _but_ the "Start" node would be null (since we tell `loadData` to empty the node list) which would throw an exception and kill the app.


src/scss/style.css
```diff
@@ -76,6 +76,8 @@ input[type='checkbox'] {
 
 #app {
   position: absolute;
+  top: 0px;
+  left: 0px;
```
TranquilMarmot, line 80: For some reason, in the VSCode webview, `#app` was a shifted few pixels to the right which was causing issues like arrows not rendering in the right spot.

Adding this gets it to the right spot, and doesn't seem to have any effect on the web app.

src/js/classes/app.js
```diff
@@ -418,6 +423,20 @@ export var App = function(name, version) {
     }
   };
 
+  // This should be called whenever we want to mark the document that the VSCcode
+  // extension has opened as changed. If we're not in the extension, this is a no-op.
+  // This should be called after every action that will result in a changed text document.
+  this.updateVsCodeExtensionDocument = function() {
```
TranquilMarmot, line 429: The extension relies on this being called to update its internal document and mark the document as "dirty" so that it can be saved.

I _think_ I got all the places that this needs to be called...
- Node added/removed
- Node dragged
- Node text editor closed

TranquilMarmot:
Okay, I think this is finished!

Check out this comment on the other PR for instructions to download/install and test it: https://github.com/YarnSpinnerTool/VSCodeExtension/pull/1#issuecomment-633774424

src/js/classes/data.js
```diff
@@ -401,7 +401,11 @@ export var data = {
   },
 
   save: function() {
-    if (self.editingPath())
+    if (window.editingVsCodeFile) {
+      // if we're editing a file in the VSCode extension, it handles
+      // saving the file on its end so we do nothing here
+      return;
+    } else if (self.editingPath())
```
blurymind, line 408: the else is redundant here, having that return there will prevent if (self.editingPath() from being evaluated anyway

blurymind:
I gave this a test and am quite impressed. Some notes are due however
-  The settings window is not accessible via the editor
- the the Night Mode settings established in yarnSpinner's settings inside vscode dont seem to work - I am constantly stuck in dark moderegardless if its ticked
- Same for markup language setting , autocreate nodes setting and possibly others

TranquilMarmot:
*Settings:*
I went back and forth on the settings window in the editor.

Originally, I had added a bunch of messaging to sync the settings when they're changed in the editor but it ended up being pretty weird; it would create a `.vscode/settings.json` file at the root of the project repo which may or may not be what the user wants.

In the end I opted for just having VSCode handle _all_ the settings and removing the options from the editor altogether.

*Night mode:*
As far as night mode goes, the extension will try and detect if your VSCode theme is "dark" and auto switch to it.

If you want to go back to not-night mode (day mode?) there's another option called "Override Dark Theme Night Mode" that you have to turn on.

If this is too confusing, it can be removed and just have the user be able to change the setting.

*Markup language:*
Good catch with the markup language setting not changing! It looks like `app.setMarkupLanguage` was missing from here:
https://github.com/YarnSpinnerTool/YarnEditor/blob/6073f32980d3b99fb04b13b51b5c1636ee30e247/src/js/classes/settings.js#L15-L20

blurymind:
Other stuff seems to be missing too - if thats where you apply it

Screenshot_2020-05-28_12-52-03

@daviddq  is this right? Should it be applied at another place for those?

src/js/classes/app.js
blurymind: We can benefit from this documentIsDirty in the electron version to add an * to the window title.

Can we perhaps refactor this so yarn in general knows when a document is dirty (when running in electron or vscode).

If not, thats fine :)

blurymind: actually can you just rename this function from 
`this.updateVsCodeExtensionDocument`
to
`this.setYarnDocumentIsDirty`

later on we can use it to handle the electron case 

@daviddq do you think it would be a good idea?

daviddq: Sounds good.

daviddq:
First of all, let me say I find this feature impressively wonderful, @TranquilMarmot. I won't be using anything different but this extension when it's fully ready. Thanks a lot.

I tested the extension some commits ago and settings were not fully working. I'll give it a try this afternoon. I find removing the "File/Settings" menu absolutely the way to go inside VSCode.I wish something similar was possible for the "search" box.

@blurymind Those settings you mention don't need to be "applied" since they are queried when needed. toogleNightMode() is needed the same way setTheme() is. Nice catch!

src/index.html
```diff
@@ -175,6 +175,10 @@
           </span>
 
           <button class="bbcode-button" onclick="app.insertEmoji()" title="Emoji">😉</button>
+          
+          <!-- ko if:vsCodeApi != null -->
+          <button class="bbcode-button" id="edit-node-in-vscode" onclick="app.editNodeInVisualStudioCodeEditor(app.editing())">🛠️ Edit in Visual Studio Code Text Editor</button>
```
blurymind, line 180: This is nice, but I hope the button is not shown when running in electron or the web app :)

TranquilMarmot: Ahh yup it shows, good catch

blurymind:
Sorry to ask again, but can you update this to the latest version of yarnEditor please? @daviddq fixed a couple of nasty ones

TranquilMarmot:
@blurymind alright, updated this and the extension PR!

blurymind:
Do you feel confident this is ready to merge? Testing it, I couldn't find any regressions to Yarn

TranquilMarmot:
I missed your comment about renaming `this.updateVsCodeExtensionDocument` to `this.setYarnDocumentIsDirty`; I'll do that later today and then this should be good to go

TranquilMarmot:
Okay @blurymind this should be good to go now!

blurymind:
Thank you for working on this :+1: 

Merged now :tada:

Issue #189, opened 2020-05-29
Ctrl+S after "Save as Yarn" not working


i1qwert:
Steps:
1. Open yarn editor
2. Type something in node
3. Menu -> Save as Yarn -> Save somewhere
4. Edit again
5. Hit Ctrl + S
6. Check yarn file and see that it hasn't changed

Expected: saving with Ctrl+S after saving as yarn

same happens after just _opening_ yarn file, editing it and trying to save with Ctrl+S

blurymind:
possibly broken recently by the input refactoring :cry: 
@daviddq  can you look into this? Sorry I didn't catch the regression at the pr. These refactorings are big

daviddq:
Ctrl+S opens the browser's "save file" dialog in the webapp and saves the HTML page.

blurymind:
> Ctrl+S opens the browser's "save file" dialog in the webapp and saves the HTML page.
> 
> Currently the shortcut to save is Ctrl+Alt+S:
> 
> ![image](https://user-images.githubusercontent.com/6022674/83252712-06dfd280-a1ac-11ea-85b4-ef8d4ce03911.png)

can we make it ctrl+S again? Thats the standard in most software and it was ctrl+s before :)

blurymind:
actually maybe both ctrl+s and ctrl+alt+s should save the file. No need to add checks if its in electron

daviddq:
a) It was Ctrl+Alt+S back in 0.2.4, way before I started working on this project:

![Captura de pantalla 2020-05-29 a las 13 11 39](https://user-images.githubusercontent.com/6022674/83254088-735bd100-a1ae-11ea-9022-33a4eb2b7c8f.png)

b) Ctrl+S won't work on webapp since it's captured by the browser.

c) Ctrl+S is already being handled. I just tested in the Electron build and worked properly.

daviddq:
I just tested on Electron-Windows 0.4.20 and none of the save shortcuts work (Ctrl+S, Ctrl+Alt+S, Ctrl+Shitf+S)

blurymind:
interesting, does windows have a different way of detecting the ctrl key?  I have no access to a windows computer

daviddq:
I don't think so. I need to find the way to debug in Electron. We already have an issue in Electron (arrows not drawing until a node is edited and saved #168) and I would like to tackle both.

blurymind:
[...]

PR #191, opened 2020-05-30
Fixes issue described in #189

daviddq:


src/js/classes/data.js
```diff
@@ -404,10 +404,10 @@ export var data = {
   },
 
   save: function() {
-    if (self.editingPath())
-      self.trySaveCurrent();
+    if (data.editingPath())
+      data.trySaveCurrent();
```
blurymind, line 408: interesting, did self not work on electron?  :)

src/js/classes/input.js
```diff
@@ -1,3 +1,5 @@
+import { FILETYPE } from './utils';
```
blurymind, line 1: surprised this worked before now, we didnt have it imported

blurymind:
Nice work :+1: 
This appears to fix it and now it behaves as expected 

Merged! :tada: 

btw I now notice the save command doesnt seem to have an icon - probably nothing to do with this pr
Screenshot_2020-05-30_13-28-13

src/js/classes/data.js
daviddq: On Electron self==window 🤷

src/js/classes/input.js
daviddq: Strangely on the webapp FILETYPE is already present in the context because _app_ already imported it I guess.

blurymind: js is weird sometimes,isnt it :smile:

src/js/classes/data.js
blurymind: would be worth for checking in other places for stuff like this tripping electron

Issue #168, opened 2020-04-26
New Compile version?


Flyingvette:
Saw that autosave was implemented, but sadly I only have the last compiled version 2.4 for windows. (Of course, I didn't really check until after it would have proved useful, haha / cry). In older versions I had been manually overwriting the files with direct pulls from github. It doesn't look like that's possible anymore.

Appreciate all the work you do on this. It's my favorite dialogue / text editor and sees lots of use.

blurymind:
@daviddq actually implemented automatic electron builds, however the script doing them is missing the step to publish the builds at https://github.com/YarnSpinnerTool/YarnEditor/releases

It would be great if we can update it so our users can get the latest electron builds from master

daviddq:
@blurymind I just noticed that the automatic Windows electron build is rotten. Probably related to the themes and webpack (not sure).

blurymind:
> @blurymind I just noticed that the automatic Windows electron build is rotten. Probably related to the themes and webpack (not sure).

Just an update on this, @daviddq fixed the rotten windows autobuild, but we still lack the step in the script to add the builds to the releases page

daviddq:
Auto release is implemented on PR #177

daviddq:
@blurymind Did you give auto release process a try?

blurymind:
@Flyingvette there is a new electorn release
https://github.com/YarnSpinnerTool/YarnEditor/releases

please test and report issues

Flyingvette:
Downloaded and installed. Works great, thank you @blurymind and @daviddq . This remains a very valuable tool that I use several times a week, if not every day.

The only issue I see vs the previous build I had is that when I open up a json file none of the connecting arrows between the text boxes appear until I enter at least one of the boxes. Then the arrows downstream of the entered box will refresh. This wasn't the case when using v0.2.4. I tried saving a json out of v0.4 and reopening, but the issue persisted. I don't see an outstanding issue for this and don't know if this is intended behavior or not.

daviddq:
It is not the intended behavior and I can't even reproduce it on my environment. @blurymind Can you reproduce this problem?

blurymind:
i think it probably has nothing to do with this issue. Can you reproduce the issue on the webapp version of yarn?

Flyingvette:
I installed this directly over my previous version and also have not restarted my machine since install. I've found other problems too, but I agree they should be separate issues.

daviddq:
[...]

PR #194, opened 2020-05-31
Fix delete warning

daviddq:
This PR fixes the issue mentioned here https://github.com/YarnSpinnerTool/YarnEditor/issues/168#issuecomment-636368297

Additionally: 

- Replaced the old delete dialog by a SweetAlert
- Performance boost when deleting a bunch of nodes and loading big files multiple times. To test this:

0) Checkout "master"
1) Launch YarnEditor webapp
2) Load "stressTest" file
3) Select all nodes ("Ctrl+A")
4) Delete them ("Supr" key)
5) Wait 😓
6) Now load again the "stress test" file and see how each time you load it takes more time. 😓😓

Now repeat the whole process with this branch.

### DIFF
--- a/src/js/classes/app.js
+++ b/src/js/classes/app.js
@@ -358,13 +358,10 @@ export var App = function(name, version) {
     window.parent.dispatchEvent(event);
   };
 
-  this.limitNodesUpdate = function (limit = false) {
-    if (limit){
-      self.nodes.extend({ rateLimit: { method: 'notifyWhenChangesStop', timeout: 250 } });
-    }
-    else {
-      self.nodes.extend({ rateLimit: { method: 'notifyAtFixedRate', timeout: 0 } });
-    }
+  this.limitNodesUpdate = function ( fn ) {
+    self.nodes.extend({ rateLimit: { method: 'notifyWhenChangesStop', timeout: 250 } });
+    fn();
+    self.nodes.limit( callback => () => callback() );
   };
 
   this.getNodesConnectedTo = function(toNode) {
@@ -558,16 +555,14 @@ export var App = function(name, version) {
 
     Promise.all(promises)
       .then( () => {
-        self.limitNodesUpdate(true);
-        {
+        self.limitNodesUpdate( () => {
           for (let i = list.length-1; i >= 0; --i)
             self.deleteNode(list[i]);
 
           self.updateNodeLinks();
           self.workspace.deselectNodes(list);
           self.workspace.updateArrows();
-        }
-        self.limitNodesUpdate(false);
+        });
       });
   };
 
--- a/src/js/classes/app.js
+++ b/src/js/classes/app.js
@@ -58,7 +58,6 @@ export var App = function(name, version) {
   this.name = ko.observable(name);
   this.version = ko.observable(version);
   this.editing = ko.observable(null);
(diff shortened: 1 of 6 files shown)


blurymind:
cool beans, thank you for refactoring this. I will give it a test at lunch hopefully and get it merged if it's got no regressions :+1:

src/js/classes/app.js
```diff
@@ -359,6 +358,15 @@ export var App = function(name, version) {
     window.parent.dispatchEvent(event);
   };
 
+  this.limitNodesUpdate = function (limit = false) {
+    if (limit){
+      self.nodes.extend({ rateLimit: { method: 'notifyWhenChangesStop', timeout: 250 } });
```
blurymind, line 363: It feels like we are adding more patchy fixes just so we can use this extend method. We didn't have it before,so now I wonder if we could get rid of it. Why do we need it to begin with and why is it worth writing all this stuff around it to prevent things from breaking or getting slower? I am not a big fan of adding more  calls to keep track of

daviddq: This piece of code is precisely what allows us now to improve the speed of loading big files and deleting a pile of nodes.

- As you know, the app data model is bound to the HTML using Knockout bindings.
- Everytime you modify the data model (add/modify/delete a node) Knockout triggers DOM modifications.
- This DOM modifications are expensive in terms of time.
- Furthermore, some YarnEditor operations, update links, drawing arrows, etc. have quadratic cost ( O(n^2) )
- Quadratic cost + expensive operations slow down YarnEditor.

My solution is to prevent Knockout to update the DOM on every operation and update only once when all the stuff has been done (loading all the file, deleted all the nodes) instead of refreshing the DOM each time a node is added/deleted.

On this PR I avoided "hacking" Knockout extender to disable the rate limiter.

blurymind: Ah yes, I noticed you removed the "Hack" line. Thanks for explaining. Ok, let me test this a bit more and will merge it

blurymind:
This PR makes deletion of many nodes much faster and also improves the UX by telling the user how many will be deleted.

The old issue of getting stuck in the nodes editor and breaking autocreation is however still in.

To trigger it, drop the stressTest file, then open the Start node and create a link to a non existent node

Peek 2020-06-01 17-25

daviddq:
🤦‍♂️

blurymind:
would be helpful if we had some tests in place to catch these :) but I guess it would add some headache with updating them every time we refactor something

blurymind:
side rant: I was at one point thinking of porting this whole app to svelte and get rid of knockout and jquery altogether. Svelte has stores that give us an easy way to control what recomputes when. But It would be a huge undertaking and maybe not everyone likes svelte and some would prefer react or something else - so that kind of made me abandon the idea  :)

daviddq:
I think I finally fixed the issue with Knockout's observable rate limiter in a non-excessively-hacky way.

About tests: it  would be great but a lot of work.

About Svelte: it would be AWESOME but a lot of work.

blurymind:
Nice work on fixing this! :+1:  Merged :tada: 

There is a minor annoyance where if you confirm to delete a node by hitting enter, yarn deletes the node, then opens it in the node editor. But its not such a big deal, I think we might have had it before

PR #195, opened 2020-06-04
Properly handles "intro" key when deleting node(s)

daviddq:
When confirming a node deletion with "intro" key the node editor is no longer opened.

Additionally, the delete confirmation is not shown if the number of nodes to be deleted is zero.


blurymind:
Thank you for the quick fix :) I will give this a test and merge it on lunch

blurymind:
Thank you again :+1: 

merged! :tada:

PR #216, opened 2020-10-06
Implement node title validation

### LINKED ISSUE #215: The node title seems to accept values which are not permitted
About a year ago I read the YarnSpinner documentation and today I'm looking for projects I can contribute to (hacktoberfest). I just downloaded the editor while having a vague idea about the syntax and I noticed the Editor doesn't seem to validate the input of the Node title. 

I've created a single node with as name "1st line" which was permitted in the editor but threw an error in Unity. This name contains a space so is invalid. 
> The name of the node, in this case nodename, needs to be unique across the entire project, and can be any combination of upper or lower case letters and numbers. Node titles can’t have spaces.

StefanJanssen95:
Fixes #215 

Besides adding the validation for the title I also optimized the jquery calls in the same function to avoid retreiving the same field multiple times. I also now remove the title attribute when it's not used, this to avoid empty tags and to make it easier to show the error in a dedicated part of the form instead of only when hovering.

### DIFF
--- a/src/js/classes/app.js
+++ b/src/js/classes/app.js
@@ -458,15 +458,19 @@ export var App = function(name, version) {
 
   this.validateTitle = function() {
     var enteredValue = document.getElementById('editorTitle').value;
+    var editorTitle = $('#editorTitle');
     if (
       self.getOtherNodeTitles().includes(enteredValue) ||
       self.titleExistsTwice(enteredValue)
     ) {
-      $('#editorTitle').css({ color: 'red' });
-      $('#editorTitle').attr('title', 'Another node has the same title');
+      editorTitle.css({ color: 'red' });
+      editorTitle.attr('title', 'Another node has the same title');
+    } else if (!RegExp('^[a-z0-9]+$', 'i').test(enteredValue)) {
+      editorTitle.css({ color: 'red' });
+      editorTitle.attr('title', 'Only upper or lower case letters and numbers are allowed in a node title.');
     } else {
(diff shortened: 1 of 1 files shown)


src/js/classes/app.js
```diff
@@ -458,15 +458,19 @@ export var App = function(name, version) {
...
       self.getOtherNodeTitles().includes(enteredValue) ||
       self.titleExistsTwice(enteredValue)
     ) {
-      $('#editorTitle').css({ color: 'red' });
-      $('#editorTitle').attr('title', 'Another node has the same title');
+      editorTitle.css({ color: 'red' });
+      editorTitle.attr('title', 'Another node has the same title');
+    } else if (!RegExp('^[a-z0-9]+$', 'i').test(enteredValue)) {
+      editorTitle.css({ color: 'red' });
+      editorTitle.attr('title', 'Only upper or lower case letters and numbers are allowed in a node title.');
```
blurymind, line 470: this is very subtle and only hovering over with a mouse will reveal it, but I will take it :)

blurymind:
Thanks for these. Both very clean PRs, valuable usability improvements

blurymind:
merged 🎉  thanks again :)

src/js/classes/app.js
StefanJanssen95: Yeah, the previous error was setup the same way, I'm currently looking if I can show the error in a text below the field, unfortunately I couldnt make it work using css ::after, for some reason it would never show the actual pseudoelement. I'm still looking into this.

PR #222, opened 2020-11-30
Close #201 - Fix alignment buttons

### LINKED ISSUE #201: Align nodes does not work on MacOS
Problem: Pressing the Align nodes says "Select nodes to align" even when having nodes selected.

App version: 0.4.44 standalone MacOS

Steps to reproduce:
Select nodes and press any align button.

align1
align2

FaultyFunctions:
This change was needed because alignment buttons would count as "clicking" on the workspace thus deselecting the nodes you wanted to align before the proper alignment functions could be called.

I removed the 
```javascript
if (!e.altKey && !e.shiftKey)
    app.workspace.deselectAll();
```
line because I couldn't figure out what this was even for? I didn't seem to have a use, and was getting in the way of the alignment buttons functioning. If you have more insight please let me know and I'll try to come up with a different solution if that line of logic was actually useful for something.

### DIFF
--- a/src/js/classes/input.js
+++ b/src/js/classes/input.js
@@ -50,17 +50,16 @@ export const Input = function(app) {
       this.isLeftButtonDown = (e.button === MouseButton.Left);
 
       if (app.inWorkspace()) {
-        if (!e.altKey && !e.shiftKey)
-          app.workspace.deselectAll();
+        if (self.isDragging) {
+          switch (e.button) {
+          case MouseButton.Left:
+            app.workspace.onMarqueeStart({ x: e.pageX, y: e.pageY });
+            break;
 
-        switch (e.button) {
-        case MouseButton.Left:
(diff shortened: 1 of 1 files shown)


JujuAdams:
👏 👏

blurymind:
I wasnt able to replicate this issue on windows, but the change seems sensible to me and doesnt appear to introduce any regressions. @desplesda you were on a mac right? Can you confirm this fixed the problem for you?

shdwcat:
FWIW, the alignment buttons didn't work for me on the web version either

blurymind:
Ok, I was just able to replicate the bug and this PR does fix it!
Merged to dev 🎉 👏 

blurymind:
> FWIW, the alignment buttons didn't work for me on the web version either

I deployed the fix to the web version, so it should become live in 5-10 minutes

PR #224, opened 2020-12-01
Add dracula theme

FaultyFunctions:
I added the Dracula theme because the night mode that was just inverting colors was bothering me. It's not perfect but it's a way big improvement IMO. (Dracula: https://draculatheme.com/).

Removal of the "night mode" might be an option now if you guys want! Or it can be kept in just in-case.

Here are some images!
image
image
image

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -154,7 +154,7 @@
     <canvas id="grid-canvas" class="grid-canvas"></canvas>
 
     <!-- arrow bg canvas for linked nodes -->
-    <canvas class="arrows"></canvas>
+    <canvas class="arrows" id="arrows"></canvas>
 
     <!-- foreach loop  of the nodes -->
     <div class="nodes">
--- a/src/js/classes/app.js
+++ b/src/js/classes/app.js
@@ -466,14 +466,14 @@ export var App = function(name, version) {
       self.getOtherNodeTitles().includes(enteredValue) ||
       self.titleExistsTwice(enteredValue)
     ) {
-      editorTitle.css({ color: 'red' });
+      editorTitle.attr('class', 'title title-error');
       editorTitle.attr('title', 'Another node has the same title');
     } else if (!RegExp('^[a-z0-9]+$', 'i').test(enteredValue)) {
-      editorTitle.css({ color: 'red' });
+      editorTitle.attr('class', 'title title-error');
       editorTitle.attr('title', 'Only upper or lower case letters and numbers are allowed in a node title.');
     } else {
-      editorTitle.css({ color: '#666' });
       editorTitle.removeAttr('title');
+      editorTitle.removeClass('title-error');
     }
   };
 
--- a/src/js/classes/app.js
+++ b/src/js/classes/app.js
@@ -30,8 +30,9 @@ export var App = function(name, version) {
 
   this.setTheme = function(name, e) {
     let themeName = e ? e.target.value : name;
-    $('#theme-stylesheet').attr('href', Utils.getPublicPath(`themes/${themeName}.css`));
     setTimeout(self.initGrid, 35);
+    setTimeout(self.workspace.updateArrows, 35);
+    $('#theme-stylesheet').attr('href', Utils.getPublicPath(`themes/${themeName}.css`));
   };
 
   this.setLanguage = function(language, e) {
--- a/src/js/classes/ui.js
+++ b/src/js/classes/ui.js
@@ -12,7 +12,8 @@ export const UI = function(app) {
   // Theme selector -----------------------------------------------------------
   this.availableThemes = [
     { id: 'classic', name: 'Classic' },
-    { id: 'blueprint', name: 'Blueprint' }
+    { id: 'blueprint', name: 'Blueprint' },
+    { id: 'dracula', name: 'Dracula' }
   ];
 
   // Playtest selector -----------------------------------------------------------
--- a/src/js/classes/workspace.js
+++ b/src/js/classes/workspace.js
@@ -442,7 +442,7 @@ export const Workspace = function(app) {
 
     // self.context.clearRect(0, 0, self.canvas.width, self.canvas.height);
     self.context.lineWidth = lineWidth;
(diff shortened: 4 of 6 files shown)


desplesda:
This is really nice, and a great improvement over the simple inverted colour! 

I'll have an opportunity to review this PR and your other(s) in the next couple of days. Thank you!

blurymind:
oh this looks groovy! 👍 🧛 
Thanks for these PRs btw. I will try to review them over the weekend

FaultyFunctions:
Yeah no worries, let me know if anything is scuffed! haha these are my first ever pull requests if I'm being honest. Thanks!

blurymind:
I had a look at this theme and it is beautiful as well as functional on smartphone screens like our other themes.

I have only a couple of minor complaints, but if they are hard to fix we can leave them for another pr.

CaptureDracula

On the bottom buttons:
- 1: Would it be possible to make them semi-transparent like they are on the other themes? Or at least do it when the screen is very narrow. It really helps with usability on smartphones.
- 2: You have this lighter box inside the undo buttons, which makes them a little bit distracting

blurymind:
There is another issue:
On the open from gist dialogue (file->open Gist), the dropdown items that are not hovered are near impossible to read

CaptureDracula2

blurymind:
Some of the file menu icons have bad contrast
CaptureDracula4

You can fix that by using a css filter on this theme, but not sure if its worth the effort. Depends on how much you want to polish this :)

FaultyFunctions:
Okay I pushed a fix for these issues, thanks for bringing them up! I think getting better icons that work for both themes would be ideal IMO. Would you guys be apposed to adding font-awesome (https://fontawesome.com/) to the project? These are free to use/open source and they are really consistent. I think replacing the file menu icons with these (and maybe event the editor icons for bold, italics, and all that stuff) would be a good option. They are a bit inconsistent right now in how they are implemented and how they look (sizes and divs and stuff). Lemme know!

blurymind:
lets do fontawesome as another pr :) 
If we are to add it as a dependency, it would be good to see if and how much it increases the distribution size of the app. Also confirm the icons look good on all themes

blurymind:
This looks perfect now 🎉 There is just one last thing - which we can fix in another pr if you like:

The Node title field has validation. For example if a Node with the title you enter already exists, the text will become red like so:
nodeTitleValidation


This doesn't work on your theme

blurymind:
after merging the grid feature pr, this one has some conflicts. Can you pull the new changes in dev to your dracula theme PR branch and resolve them? Then push to it again

FaultyFunctions:
Okay it should be good now I think!

FaultyFunctions:
image

Hey I have a question, is there a reason you can't enter other node titles in your node title and also is there a reason why punctuation like periods and underscores aren't allowed in the regex? Is this a Yarn specific thing?

blurymind:
Yes both of those validations are there to prevent the yarn parser from
crashing (we use bondagejs). You can playtest yarns inside the editor if
you want to see bondagejs in action

blurymind:
Yarn can also autocreate new node when you navigate to a link that has no
node yet. The way it does it is by checking if the node you closed already
exists by name  or not. If it doesn't, yarn makes a new node, if it does it
saves over the node.

It shouldn't be allowed for two nodes with the same name to exist in a yarn
due to the way the story parser works anyways

desplesda:
You can absolutely have two nodes that share a prefix in Yarn Spinner - I suspect that the use of `includes` here is not quite what this code is intended to do. Nodes can't share a name, but otherwise there's no significant restrictions besides no whitespace.

blurymind:
Unfortunately yarn spinner works only in unity and yarn editor itself can
not use it. This is why we use bondagejs. Bondagejs is also used in a
number of html5 game engines such as gdevelop and ctjs.

You can of course add support for that to bondagejs if you want to. Its got
its own tracker :)

blurymind:
Oh if its just the name prefix then bondage has no problem with it. Yarn
editor should allow two nodes with the same prefix to exist. If it doesn't
its a bug

FaultyFunctions:
OH I had a misunderstanding. It allows prefixes but since the regex that is currently in the code doesn't allow anything other than letters and numbers when I was typing an underscore I was getting an error. Because my nodes were titled "Node_1", "Node_2", etc... I thought it was a prefix error for "Node_" but it's actually the underscore that isn't passing the regex check.

So what characters *should* be allowed then? Is there any sort of specification for this for the Yarn language in general?

blurymind:
Well ideally both should have identical rules but as I said YarnSpinner and Bondagejs (the thing Yarn editor uses for playtesting) have some subtle differences. Bondagejs does not allow some things that spinner allows and thats something that needs fixing with bondagejs.

In any case this warning colour will not stop the user from creating nodes - it will only warn you that the name is going to crash the playtest feature.

If you are to fix that, I think it should be for a separate pr and this one should only be about the new awesome dracula theme.
We should really continue the discussion on an issue ticket- it will get lost once this pr is merged in

blurymind:
actually if you can test if what the matched regex allows doesn't crash yarn editor when you playtest yarn files with such titles and confirm that it doesnt - can you do the changed regex as a separate pr. 
https://github.com/YarnSpinnerTool/YarnSpinner/blob/483f586b752f93000ea867cbad32a28eb55e4359/YarnSpinner.Compiler/Compiler.cs#L113
It will make it easier to test :)

FaultyFunctions:
I agree! I was getting a bit ahead of myself tbh, I will do the tests once I get past the other stuff I'm working on right now. The theme should be ready to merge now, thanks for all the help and feedback guys!!

blurymind:
Thanks for working on these awesome new features and fixes. Your contributions will be felt by the Gdevelop community as well when we get the yarnEditor that is bundled with it updated.

I will have another look tonight and get this merged if there are no issues 👍

blurymind:
Merged! 🎉 
Thank you for working on this. This is Yarn's very first dark theme and it looks beautiful 🕶️ 👨‍🎨 🎨

Issue #147, opened 2020-04-08
Align nodes vertically and horizontally


daviddq:
Currently there's an option to _arrange_  the nodes vertically and horizontally:

![Captura de pantalla 2020-04-08 a las 15 33 47](https://user-images.githubusercontent.com/6022674/78790237-845a4400-79ae-11ea-83c8-c5f0c868be45.png)

I find it unusable since both change X and Y of the nodes. Starting with this state:

![image](https://user-images.githubusercontent.com/6022674/78790366-b53a7900-79ae-11ea-8811-89b84543c56e.png)

when aligning vertically I would expect this (only X components are aligned):

![image](https://user-images.githubusercontent.com/6022674/78790453-d13e1a80-79ae-11ea-9fbc-b7331c4fc947.png)

but I get this (X components are aligned and Y have been _reduced_):

![image](https://user-images.githubusercontent.com/6022674/78790472-dc914600-79ae-11ea-8dfe-bb5cad8d0b2f.png)

Similar thing happens when aligning horizontally.

I would like:

a) That the tool works the way I expect and explained (prefered)

or

b) To have a more complete menu where I can choose between "arrange" (current behavior), "align" (my proposal), "distribute space", etc.

blurymind:
I see what you mean, it changes the spacing between them, it doesnt just align them.
A pr to amend it is welcome :) 

Btw this feature is quite a legacy thing. I kind of revived it and broke a bug that prevented it from working a while ago

I want to keep it simple if possible.

How about:
1. Single click/tap the button does the behaviour you want
2. Double click/tap the button does the old behaviour where they are arranged with equal spacing between them

daviddq:
I also prefer option 2) to having a more complex menu 👍 

What about using `ctrl+click` or `alt+click` for option 2) instead of double clicking ?

blurymind:
> I also prefer option 2) to having a more complex menu 
> 
> What about using `ctrl+click` or `alt+click` for option 2) instead of double clicking ?

I'm alright with that too :)
Only reason I suggested doubleclicking was mobile devices

FaultyFunctions:
So is this still being worked on? Those alignment buttons don't work at all for me, they just keep saying select nodes to align.

PR #225, opened 2020-12-03
Feat align double click

FaultyFunctions:
Single click will align H or V but not reduce and double click or ctrl click will align H or V AND reduce (the previous behavior) as outlined in #147.

This might be a temp fix since you guys were discussing a more complex menu but I didn't want to mess with having a drop down when you click on the alignment buttons, so this is a good middle ground that you guys had discussed.

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -59,8 +59,8 @@
 
     <!-- align controls -->
     <div class="app-sort">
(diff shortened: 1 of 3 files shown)


FaultyFunctions:
Oh snap I accidentally added the dracula stuff to this branch. Hold up.

PR #226, opened 2020-12-03
Close #147 - Double click or ctrl click for reduce align functions

### LINKED ISSUE #147: Align nodes vertically and horizontally

FaultyFunctions:
- Single click on alignment buttons will align the x or y but won't reduce space
- Double click or control click on alignment buttons will both align H and V and reduce the other axis (this is the previous behavior)
- Renamed align functions to be less confusing

Okay this is the fixed version! My B!

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -59,8 +59,8 @@
 
     <!-- align controls -->
     <div class="app-sort">
-      <span data-bind="click: app.workspace.alignX" title="Arrange H"></span>
-      <span data-bind="click: app.workspace.alignY" title="Arrange V"></span>
+      <span data-bind="click: app.workspace.alignH, event: { dblclick: app.workspace.reduceAlignH }" title="Arrange H"></span>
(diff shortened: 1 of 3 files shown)


blurymind:
oh this is very useful! Thank you for fixing it. I will have a look after work :)

blurymind:
Code looks good to me and addresses the problem cleanly. Nicely done 👍

blurymind:
Merged! 🎉 

PR #227, opened 2020-12-03
Add bezier curve option for connecting lines

FaultyFunctions:
This was just to make the connecting lines look a bit fancier and it works really well with the new grid options. I have set it as the default value but if you wish the straight lines to be the default let me know. Thanks!

### DIFF
--- a/src/js/classes/settings.js
+++ b/src/js/classes/settings.js
@@ -148,7 +148,7 @@ export const Settings = function(app) {
 
   // Line Style
   this.lineStyle = ko
-  .observable(storage.getItem('lineStyle') || 'Bezier')
+  .observable(storage.getItem('lineStyle') || 'straight')
   .extend({ persist:'lineStyle' });
 
   // Always open nodes in Visual Studio Code Editor
--- a/src/js/classes/settings.js
+++ b/src/js/classes/settings.js
@@ -141,10 +141,15 @@ export const Settings = function(app) {
     .observable(storage.getItem('markupLanguage') || 'bbcode')
     .extend({ persist:'markupLanguage' });
 
-  // Markup language
+  // Playtest style
   this.playtestStyle = ko
     .observable(storage.getItem('playtestStyle') || 'chat')
     .extend({ persist:'playtestStyle' });
+
+  // Line Style
+  this.lineStyle = ko
+  .observable(storage.getItem('lineStyle') || 'Bezier')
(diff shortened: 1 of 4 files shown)


blurymind:
Can you add a screenshot please :) sounds good!

FaultyFunctions:
Sure!
Bezier Curves:
image

Straight Lines:
image

src/js/classes/settings.js
```diff
@@ -141,11 +141,16 @@ export const Settings = function(app) {
...
 
-  // Markup language
+  // Playtest style
   this.playtestStyle = ko
     .observable(storage.getItem('playtestStyle') || 'chat')
     .extend({ persist:'playtestStyle' });
 
+  // Line Style
+  this.lineStyle = ko
+  .observable(storage.getItem('lineStyle') || 'Bezier')
```
blurymind, line 151: can we make straight the default? They read better when you have loads and loads of nodes, its also less intrusive of a change to existing users

src/js/classes/workspace.js
```diff
@@ -448,51 +448,132 @@ export const Workspace = function(app) {
...
+              x1: to.x + normal.x * arrowLength,
+              y1: to.y + normal.y * arrowLength,
+              x2: to.x - normal.x * arrowWidth - normal.y * arrowHeight,
+              y2: to.y - normal.y * arrowWidth + normal.x * arrowHeight,
+              x3: to.x - normal.x * arrowWidth + normal.y * arrowHeight,
+              y3: to.y - normal.y * arrowWidth - normal.x * arrowHeight
+            });
+          }
+        }
+      } else {
```
blurymind, line 499: why not simply wrap both instead like so
if (node.linkedTo().length) {
    if (app.settings.lineStyle() === "straight") {
       ....
    } else {
       ....
    }
}

blurymind:
Code looks good 👍 again very clean PR. I could only see a couple of minor things :)

src/js/classes/workspace.js
FaultyFunctions: Good catch! Updating now.

FaultyFunctions:
Okay should be good now!

blurymind:
Thanks for updating this. Will have another look tonight :)

blurymind:
Merged! 🎉 
Thank you for working on this. Looks nice :)

PR #228, opened 2020-12-03
Font awesome icons & design overhaul

FaultyFunctions:
Hey! I've updated the interface to look a lot cleaner IMO. Icons are a lot more consistent now and I've moved all the toggle buttons from the editor to a bottom toolbar. I also moved the text to speech, search, and play buttons to the right of the top toolbar. Also there is now some logical separators between bbcode sections in the toolbar. You can see the changes here:

image

As well as updates to the canvas zoom, add, redo/undo, and sort buttons:

image

Aaaand classic theme:
image


blurymind:
This looks cool! 
One concern looking at the screenshot is that you have reduced the vertical space for the actual text area, also changed the layout. It might be a problem on smartphones. I will test this tonight. Please remember that Yarn is optimized to work on tiny screens in portrait mode, but when android's virtual keyboard shows up, it takes half that screen and suddenly vertical real estate becomes incredibly sparse

FaultyFunctions:
Ah yes good point, let me know how it is on mobile. I can adjust vertical spacing and figure how to go about designing in that space! Thanks!

FaultyFunctions:
I made some adjustments and loaded it up on my phone, here is what it looks like on a Galaxy S9:





What do we think? Would it be possible to hide the bottom bar if the virtual keyboard is up, or is that not really do-able?

blurymind:
In order to see how many text rows it fits, hit enter until the left side gutter shows count from top to bottom.
The old layout used only one row for the top toolbar (-1 row vertical space lost)
It also did not have a bottom toolbar and the character counter was hidden by default on the small screen (I believe it was using a media query to hide some stuff)( - 2 rows vertical space lost)

Can you load the PWA on your phone and compare it with the PR - it will be easier to see what it was doing to preserve vertical space on smartphones

You also lost a feature - the app in any resolution would let you toggle the visibility of the counter (which is now always visible on the PR and is moved to the bottom)

Can we not split the toolbar into top and bottom and just shrink the buttons a bit to get them to fit as they did before? Also why the need to rearrange the layout, couldnt that be done after first talking about it in an issue - what are the benefits of this rearrangement and why? Such a drastic change should be done in another pr imo

FaultyFunctions:
Sure here are the two side by side:

image
Left being the old, right being the new. It looks like we're losing maybe 0.5 - 1 lines at max.
You actually can toggle the counter on and off with the left most button on the bottom toolbar.

The reason I changed up the layout was because the old layout mixed BBCode buttons with toggle buttons for auto complete and others and there wasn't any logical order to them. The way they are grouped in this PR makes the most sense to me, but I am open for discussion on changing things. I just think the old layout made no sense coming from any other BBCode editor. Sure it's a change people will have to get used to, but I think in the long run this is much more navigable than the old layout. They are grouped so the actual BBCode buttons are top-left, toggles are on the bottom bar, and the other buttons (transcribe, find, & play) are top right.

Let me know your thoughts and any other issues, again I'm open to discussion!

blurymind:
One more thing worth confirming also 

go on master and do
npm run build
note the dist folder's size

then go to this PRs branch, delete the contents of dist and repeat.

What is the difference between the dist folder's size here and on master?

FaultyFunctions:
Master - 4.49 MB
PR - 5.35 MB
Diff - 0.86 MB

Hmm not sure if this is bad or not. Are we trying to keep it under a certain threshold, I suppose the larger it is the slower it loads, lemme know if it's a problem and what we can do about it if it is. Thanks!

FaultyFunctions:
I'm in the process of defining the icons in the HTML file itself instead of loading the entire font-awesome library. Which I think should reduce even the master's file size since we no longer have the old image icons in there but we shall see. This isn't a problem to define the icon SVG's in the HTML file and use them that way, right?

blurymind:
Can you try to bring down the dist size? I try to keep it low as its one of the deciding points in wheter yarn should be bundled with an app.
The older version used to be around 2mb, but we somehow made the thing grow. This might be a problem when I want to upgrade the version that gdevelop is using - since I need it to be a little bit under 1mb when zipped 
https://github.com/4ian/GDevelop/releases/download/v5.0.0-beta80/yarn-editor.zip
Otherwise Florian might not allow it for too big of a size

blurymind:
Ok having a closer look, I think this can work, but has a few things that will make it as good and maybe even better.
Set your resolution to 386x386 - I tested this on my galaxy note phone and thats what I get when the virtual keyboard shows up.
I get almost 13 lines
tinyResolutionTest

- 1: Using the media query can you hide the buttons that are really not useful on smartphones. Some of these features are built  into the virtual keyboard anyways. Hide that bottom bar and some of the top toolbar buttons
- 2: Can you get the top toolbar buttons to be a single row when the resolution is 386 px wide
- 3: ignore that for now, but if you can fix ace editor not properly stretching to the bottom, that would be a bonus

FaultyFunctions:
Awesome suggestions, I'll get to implementing these!

blurymind:
I think maybe you can even hide the emoji button on smartphones,  since the
virtual keyboard has that built in. Its rarely used too

FaultyFunctions:
Alright! Quite a big commit I admit, but there is some good stuff in there.

I figured out how to hide the toolbar and tags row when you scroll down on mobile and have it reappear on scroll up.
image


Here is the regular comparison for the vertical spacing changes on the toolbar without the toolbar hidden:
image

Plus I got the dist down to 4.24 MB which is down 0.25 MB from the master branch since I removed the image icons we had stored.

Give it a test run and let me know if I missed anything or if anything can be improved/changed. Thanks!!

blurymind:
this does look pretty awesome! :) I will have a look after work today

src/js/classes/app.js
```diff
@@ -825,6 +826,30 @@ export var App = function(name, version) {
...
+      } else {
+        $('.bbcode-toolbar').removeClass('hidden');
+        $('#editorTags').removeClass('hidden');
+      }
+      self.editorPreviousScrollPosition = position;
+    });
+
+    /// Recover editor toolbars from hidden state if window is resized
+    $(window).on('resize', function() {
+      if ($(window).width() > 600) {
```
blurymind, line 847: can we store this width > 600 in a boolean variable instead?  (isOnNarrowScreen?)
Also wouldnt this whole logic make more sense to add to ui.js?

I wonder if it should be done via css

src/js/classes/settings.js
```diff
@@ -104,11 +104,11 @@ export const Settings = function(app) {
     ).extend({ persist:'completeClosingCharacters' });
 
   // Night mode
-  this.nightModeEnabled = ko
-    .observable(storage.getItem('nightModeEnabled') !== null ?
-      storage.getItem('nightModeEnabled') === 'true' :
+  this.invertColorsEnabled = ko
```
blurymind, line 107: 👍 nice. Name makes more sense now

src/js/classes/ui.js
```diff
@@ -58,6 +58,12 @@ export const UI = function(app) {
     { name: 'Lingua latīna', id: 'la' }
   ];
 
+    // Line Style selector -----------------------------------------------------------
+    this.availableLineStyles = [
```
blurymind, line 62: Wasnt this a part of another PR that got merged?
why is it appearing here?

src/js/classes/workspace.js
```diff
@@ -449,50 +449,129 @@ export const Workspace = function(app) {
...
-            x1: to.x + normal.x * arrowLength,
-            y1: to.y + normal.y * arrowLength,
-            x2: to.x - normal.x * arrowWidth - normal.y * arrowHeight,
-            y2: to.y - normal.y * arrowWidth + normal.x * arrowHeight,
-            x3: to.x - normal.x * arrowWidth + normal.y * arrowHeight,
-            y3: to.y - normal.y * arrowWidth - normal.x * arrowHeight
-          });
+            ) * scale;
+  
+            const from = {
```
blurymind, line 478: why is this appearing on this dif, after I merged the PR with the curvy lines?
Try merging dev into this branch - wonder if that would clean up the dif

src/js/classes/app.js
```diff
@@ -825,6 +826,30 @@ export var App = function(name, version) {
...
+        $('.bbcode-toolbar').removeClass('hidden');
+        $('#editorTags').removeClass('hidden');
```
blurymind, line 849: instead of hiding it, can you make scrollout of frame - the way some other web apps do?
Similar example here
https://www.w3schools.com/howto/howto_js_navbar_hide_scroll.asp

notice the transition in the css

blurymind (commented):
I added a few minor notes, but otherwise this looks pretty darn awesome and seems to work well on mobile! 👍

src/js/classes/workspace.js
FaultyFunctions: I'm not sure, I think it's showing up because I have been rebasing this branch onto master to keep it updated. So the updated dif is showing up on this branch? Not sure...

image

Should I not be rebasing?

src/js/classes/app.js
FaultyFunctions: Yeah I initially wanted to do that as well but everything I looked up was for a nav bar stuck to the top of the page that would just push it out of frame like the one you linked. Not sure how to get it working with the current HTML setup but maybe if I can put the toolbar inside the "editor-container"?? I'll give it a shot.

FaultyFunctions:
Hey unfortunately the scroll to hide the toolbar was a lot buggier than I originally thought it was so I removed it in the latest commit. I think it would be best to open up another PR for that functionality once I get more time to figure out the issues. Let me know if there needs to be any more last minute changes. Thanks!

blurymind:
> Hey unfortunately the scroll to hide the toolbar was a lot buggier than I originally thought it was so I removed it in the latest commit. I think it would be best to open up another PR for that functionality once I get more time to figure out the issues. Let me know if there needs to be any more last minute changes. Thanks!

Let's leave it for another time then :) thank you for having a try with it - I thought that it might be tricky

I will have a another look at this pr tonight and get it merged if theres nothing broken 👍

src/js/classes/workspace.js
blurymind: could be that, I am not sure actually. Its very strange

webpack.config.js
```diff
@@ -123,10 +123,6 @@ const config = {
       display: 'fullscreen',
       crossorigin: 'use-credentials', //can be null, use-credentials or anonymous
       icons: [
-        {
-          src: path.resolve('src/public/icon.png'),
```
blurymind, line 127: removing this might actually create problems for the pwa - your browser will deem it unfit to be installable. Can you revert changes you made to this file and deleting src/public/icon.png ?

blurymind: Yarn is currently an installable pwa. To be that it requires an icon
https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Installable_PWAs

so what does installable mean to a mobile user
- when you visit its web hosted location, you are offered the option to add it to your homescreen
-  when you add it to the phones homescreen and run it from there it can run even when you are offline. It also runs in full screen (the address bar is hidden)

src/js/classes/workspace.js
blurymind: when you rebased it, did you by any chance rebase it onto your  fork of the master branch (and that fork is perhaps behind this master branch)?

FaultyFunctions:
I couldn't find any spec to say what type of images are accepted so I just added the line back and pushed that. Thanks for the info though!

FaultyFunctions:
Alright I moved the icon definitions to their own folder. It was really easy too, great suggestion! Removes a lot of clutter from the main index.html.

blurymind:
thanks :)

blurymind:
Merged! 🎉 
Thank you for working on this 😄  Yarn is starting to look super sleek with your changes!

Issue #219, opened 2020-10-09
[Windows] Opening a new file won't let you edit it if you're already in another yarn file


DavidNightcap:
This is a quality of life bug. The workaround currently is to close Yarn Editor entirely and reopen which is a bit annoying, but not the end of the world.

Repro steps:
1) Open a yarn file...may or may not need to make changes.
2) Use File -> Open to open a different yarn file.
3) Try to edit text in a node...notice you can't type anything.

DavidNightcap:
I just noticed this problem is actually more annoying now with the latest Editor update because it opens the last file you had open. So now you have to switch files, close it, reopen it, to be able to edit.

Issue #218, opened 2020-10-09
[Windows] CTRL+S does not save while in node editing mode


DavidNightcap:
Not sure if this is a feature request or a bug but it's bitten me many times trying to iterate on my dialogue and testing it out in game.

Repro steps:
1) Open a yarn file
2) Open a node for editing and type stuff.
3) Hit CTRL+S to save.
4) Either test in game or open in a text editor, but your changes are not saved.

Issue #37, opened 2019-09-18
[Windows] CTRL+Z clears a node's contents


desplesda:
<a href="https://github.com/renaudbedard"><img src="https://avatars1.githubusercontent.com/u/251134?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [renaudbedard](https://github.com/renaudbedard)**
_Monday Aug 22, 2016 at 15:07 GMT_
_Originally opened as https://github.com/InfiniteAmmoInc/Yarn/issues/37_

----

Test case : 
- Create a node, add content
- Exit node
- Double-click on that same node
- Hit CTRL-Z
- Node text disappears, and that operation can't be undone

Before hitting CTRL+Z

![image](https://cloud.githubusercontent.com/assets/251134/17859719/bce243ec-6858-11e6-8747-acfb02dc122c.png)

After hitting CTRL+Z

![image](https://cloud.githubusercontent.com/assets/251134/17859725/c7cf149c-6858-11e6-86e7-93be05c2105d.png)

desplesda:
<a href="https://github.com/rcuddy"><img src="https://avatars2.githubusercontent.com/u/4494839?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [rcuddy](https://github.com/rcuddy)**
_Wednesday Jun 28, 2017 at 10:28 GMT_

----

This 'clear' action on ctrl + z still happens, but at least on 0.2.1 ctrl + y puts the text back.

From what I can tell in the debugger the bit that's causing this to happen is ace.js undoChanges function. It calculates a delta of changes and if I understand it correctly, the act of loading a node into the ace editor causes a delta state of "it was empty, then I had a bunch of text". The wierd set of tabs and spaces you get are the result of the tabs inside the index.html `<div class="editor" id="editor" .... >{tab and spaces}</div>` which  ace considers the initial state to fall back to. If you remove those spaces, or populate the empty space with your own text, you can see that when the undo delta triggers you get whatver was between the open and closing tags.

That's great, that's the problem identified.

I have no idea how ace works though, so I'm not sure how to fix this. Does anyone have any suggestions on how we could interrupt the ace undo method if we're at the first local change which was "load node text into editor" ? I though the ctrl+z stuff in app.js updateHighlights may have been what was supposed to be doing this, but it seems that block of code is never actually called, at least that I've seen as yet.

PR #237, opened 2021-01-19
Various bug fixes & improvements

FaultyFunctions:
Full changes:

Editor:
- Rename `Complete Tags` to `Auto Close Tags` where appropriate.
- Rename `Complete Words` to `Autocomplete Suggestions` where appropriate.
- Rename `Complete Closing Characters` to `Auto Close Brackets` where appropriate.
- Fix bug where the three tags mentioned above wouldn't toggle properly.
- Fix how the Auto Close Tags works and remove unnecessary spaces.
- Removed the auto close tag for `[[ answer: | ]]`.  Not every one follows this format in their Yarn files and this will be depreciated anyway in the 2.0 Yarn spec. Same thing when clicking `Add Link` button.
- Increased `Add Link` font a little to be more readable.
- Added panning to selected node when you open the split editor.
- Updated editor title style to accommodate really long node titles for desktop & mobile.
- Split editor will now remember its previous width.

Canvas:
- Hopefully fixed a bug where grid color didn't register sometimes? Need to find a better way to update this rather than setting a timeout.
- Convert alignment alerts to Swal notifications to be less intrusive.
- Fix bug where panning to nodes would break the canvas if not in-editor and split-editor is enabled.

Dialogs:
- New file dialog will now ask for a file name after you confirm you want to open a new file.
- Editing the new file dialog a bit to better match the rest of the app's dialog aesthetics.

Try it out and let me know if you find any problems! Thanks!


FaultyFunctions:
Also once this is merged and taken care of, I was wondering if we could get a new release so people can download the new changes? Thanks!

FaultyFunctions:
Oops I maybe broke opening files on the web version. Investigating now...

FaultyFunctions:
https://github.com/YarnSpinnerTool/YarnEditor/commit/000ab57ace52cac255e4987ee09be35862b3b6a1 closes #219

blurymind:
> Also once this is merged and taken care of, I was wondering if we could get a new release so people can download the new changes? Thanks!

Releases stopped coming because the electron automatic builds stopped working. I welcome a pr to fix them. Maybe @daviddq can help?

blurymind:
I am against removing the autoclose tag for [[ or anything deprecating the old spec.
It is the only way to make a linked node for bondagejs users. Please keep compatibility with the old yarn until we can spin spec2 with a js module.

FaultyFunctions:
Oh sorry there has been a misunderstanding. I just removed the *full* autocompletion of it completing `[[` to `[[ answer: | ]]` since not everyone uses the full format and some people want to use just `[[node]]`. This doesn't remove support for the `[[` tag at all, it merely just doesn't pick one format over the other. This makes it less annoying for people who still want to use auto close tags for `[b][/b]` but don't want to get `[[ answer: | ]]` every time they type `[[`, if you understand what I mean.

blurymind:
Ah phew you gave me a scare for a moment. I am fine with that then :)

FaultyFunctions:
> > Also once this is merged and taken care of, I was wondering if we could get a new release so people can download the new changes? Thanks!
> 
> Releases stopped coming because the electron automatic builds stopped working. I welcome a pr to fix them. Maybe @daviddq can help?

I have no idea how the auto build stuff works or the dependabot thing functions. So any insight into that would be appreciated.

FaultyFunctions:
c58fe89 fixes #218

FaultyFunctions:
With d9fcd25 this can open up the possibility in the future of saving the entirety of the undo history in a file between sessions. So if you save a file and exit the editor, then open it back up again, you can still press Ctrl+Z and it'll undo text like you had just typed it. However, that's future work for a later time!

daviddq:
> > Also once this is merged and taken care of, I was wondering if we could get a new release so people can download the new changes? Thanks!
> 
> Releases stopped coming because the electron automatic builds stopped working. I welcome a pr to fix them. Maybe @daviddq can help?

I will take a look next weekend.

blurymind:
> With [d9fcd25](https://github.com/YarnSpinnerTool/YarnEditor/commit/d9fcd2571d16869751f7e11afa5b52b465d9e244) this can open up the possibility in the future of saving the entirety of the undo history in a file between sessions. So if you save a file and exit the editor, then open it back up again, you can still press Ctrl+Z and it'll undo text like you had just typed it. However, that's future work for a later time!

cool :) it looks like the beginning of an undo manager. Would the file history be stored inside the file or another hidden file mirroring it? How will that work with gist hosted files?

Anyways, please keep big features like this for a separate PR. In fact if this pr is not using it to fix anything in any way, I would recommend removing it from the pr.

FaultyFunctions:
Oh this data isn't stored in between sessions anywhere yet. It doesn't affect anything currently, it just keeps undo history internally while moving in between nodes. Once you exit the editor the history is reset just like it is currently. The way this works is by adding `this.undoManager = null;` variable inside `node.js`. Then when you exit the currently edited node it saves Ace's `UndoManager` object inside this variable, then when you load the node back up for editing it re-applies this `UndoManager` object to the editor.

That comment is just speculation on what we can do with this in the future, otherwise this feature is main just to fix #37.

We can open another PR later or issue to discuss how this should work in full if we do decide to tackle actually saving undo history on the file.

blurymind:
> Oh this data isn't stored in between sessions anywhere yet. It doesn't affect anything currently, it just keeps undo history internally while moving in between nodes. Once you exit the editor the history is reset just like it is currently. The way this works is by adding `this.undoManager = null;` variable inside `node.js`. Then when you exit the currently edited node it saves Ace's `UndoManager` object inside this variable, then when you load the node back up for editing it re-applies this `UndoManager` object to the editor.
> 
> That comment about saving the history was just speculation on what we can do with this in the future, otherwise this feature is mainly just to fix #37.
> 
> We can open another PR later or issue to discuss how this should work in full if we do decide to tackle actually saving undo history on the file.

ah clever - I like that you hook directly into the undo manager :)

Can you put all of  d9fcd25 in another PR once its more mature? It will make it much clearer to read on the diff of that pr.

FaultyFunctions:
Yeah sure but how would that work? I'm not sure how to accomplish that.

blurymind:
> Yeah sure but how would that work? I'm not sure how to accomplish that.

revert the commit on the branch of this pr (various-bug-fixes) and push it to another branch you have created? (maybe implement-undo-manager)

electron/package.json
```diff
@@ -40,6 +40,15 @@
     ]
   },
   "build": {
-    "productName": "Yarn Editor"
+    "productName": "Yarn Editor",
+    "fileAssociations": {
+      "ext": "yarn",
```
blurymind, line 45: ah, thats very nice :)  will let the electron version open yarn files via the OS file manager?

FaultyFunctions: After some research it seems only Windows 10 Universal Apps and Chrome Apps have a file handling API for this. This is being worked on I think for regular PWA apps?
See:
https://www.reddit.com/r/PWA/comments/gol8jh/open_with_pwa/

Not sure if this is possible or not with out current set up.

src/index.html
```diff
@@ -52,18 +52,18 @@
     </div>	
     <!-- undo/redo controls -->	
     <div class="app-undo-redo app-button">	
-      <span data-bind="click: function() { app.historyDirection('undo'); }">
+      <span data-bind="click: function() { app.historyDirection('undo'); }" title="Undo">
```
blurymind, line 55: nice :) I appreciate the tooltips

src/js/classes/app.js
```diff
@@ -374,7 +383,10 @@ export var App = function(name, version) {
     this.hearText = function() {
       const available = spoken.listen.available();
       if (!available) {
-        alert('Speech recognition not avaiilable!');
+        Swal.fire({
```
blurymind, line 386: nice, thanks for getting rid of the alerts

src/js/classes/app.js
```diff
@@ -17,6 +17,7 @@ import { data } from './data';
 import { yarnRender } from './renderer';
 import { Utils, FILETYPE } from './utils';
 import { RichTextFormatter } from './richTextFormatter';
+import Swal from 'sweetalert2';
```
blurymind, line 20: Iwonnder if this would work without importing it here. We are already importing it in index.js 🤔

blurymind:
thanks for this, the code looks alright 👍  just need to test it for regressions

electron/package.json
blurymind: There is a new api that gives access to the file system now
https://web.dev/file-system-access/
but there is a second part to this - how do we get the path to the file that is opened with the installed pwa? 

This one might be a can of worms for another pr i guess :)

electron/yarn-main.js
```diff
@@ -15,7 +18,19 @@ window.addEventListener("yarnReady", e => {
 });
 editorFrameEl.src = "app/index.html";
 
+window.addEventListener("yarnReadyToLoad", e => {
+  yarn = e;
+  // Open file on doubleclick on Windows
+  if (remote.process.platform === 'win32' && remote.process.argv.length >= 2 && !isDev) {
```
blurymind, line 24: this is a bit flaky. electron has a better way of dealing with this,
https://www.electronjs.org/docs/api/shell#shellopenpathpath

also if you are only checking for win32, what about other OS? You need to be checkingn if it is running on electron instead

src/js/classes/data.js
```diff
@@ -126,23 +142,50 @@ export const data = {
...
-    }
-    data.editingName(filename.replace(/^.*[\\\/]/, ''));
-    data.readFile(file, filename, true);
-    data.isDocumentDirty(false);
-    data.editingPath(file.path);
-    data.lastStorageHost('LOCAL');
-    app.refreshWindowTitle();
+    })
+  },
+  openFileOnStart: function(path, filename) {
```
blurymind, line 166: call this openFileFromLocalPath. A more descriptive and generic name is better imo

electron/yarn-main.js
```diff
@@ -15,7 +18,19 @@ window.addEventListener("yarnReady", e => {
 });
 editorFrameEl.src = "app/index.html";
 
+window.addEventListener("yarnReadyToLoad", e => {
```
blurymind, line 21: instead oh kicking this off with an event listener, you can do  it in electron/main.js

src/js/classes/app.js
```diff
@@ -144,7 +147,13 @@ export var App = function(name, version) {
       return null;
     });
     window.addEventListener('DOMContentLoaded', e => {
-      this.data.loadAppStateFromLocalStorage();
+      if (self.isElectron) {
+        var event = new CustomEvent('yarnReadyToLoad');
```
blurymind, line 151: I believe we aready have something to detect this event

see
https://github.com/4ian/GDevelop/blob/ce5282f41b6ff0eafbd3063b20237e88d9262596/newIDE/app/public/external/yarn/yarn-main.js#L24

src/js/classes/data.js
```diff
@@ -126,23 +142,50 @@ export const data = {
...
+      url: path,
+      async: false,
+      success: (result) => {
+        let type = data.getFileType(filename);
+        if (type == FILETYPE.UNKNOWN) {
+          Swal.fire({
+            title: 'Unknown filetype!',
+            icon: 'error'
+          });
+          self.loadAppStateFromLocalStorage();
```
blurymind, line 177: if it fails to load the file on start and loads a previous one - would that be a bit confusing? The user will think it loaded a different file than the one they double clicked on

src/js/classes/data.js
```diff
@@ -126,23 +142,50 @@ export const data = {
...
+        } else {
+          data.editingName(filename.replace(/^.*[\\\/]/, ''));
+          data.editingPath(path);
```
blurymind, line 180: I wonder if you can use data.openFile instead of replicating it here

electron/package.json
```diff
@@ -40,6 +40,15 @@
...
+      "name": "Yarn File",
+      "role": "editor"
+    },
+    "win": {
```
blurymind, line 49: can you add icon for mac and linux too?

src/js/classes/data.js
```diff
@@ -1,6 +1,7 @@
 /* eslint-disable jquery/no-ajax */
 const path = require('path');
 const saveAs = require('file-saver');
+import Swal from 'sweetalert2';
```
blurymind, line 4: again, I dont think you need to import it here :)

src/js/classes/richTextFormatterHtml.js
```diff
@@ -1,15 +1,15 @@
...
-    { Start: '[[', Completion: ' answer: | ]] ', Offset: -4 },
-    { Start: '<<', Completion: ' >> ', Offset: -3 },
-    { Start: '<colo', Completion: 'r=#></color> ', Offset: -10, Func: () => { app.insertColorCode(); } },
-    { Start: '<b', Completion: '></b> ', Offset: -5 },
-    { Start: '<img', Completion: '></img> ', Offset: -7 },
-    { Start: '<i', Completion: '></i> ', Offset: -5 },
-    { Start: '<u', Completion: '></u> ', Offset: -5 },
-    { Start: '<url', Completion: '></url> ', Offset: -7 },
+    { Start: '<<', Completion: '>>', Offset: -2 },
+    { Start: '<colo', Completion: 'r=#></color>', Offset: -9, Func: () => { app.insertColorCode(); } },
```
blurymind, line 7: if you right click on the hex code of a colour, and select a different one - does that properly replace the hex code?

electron/yarn-main.js
FaultyFunctions: Running it on dev makes `remote.process.argv` return an array with like 20 entries that doesn't contain any information we care about so without the isDev this would trigger and error out.

I wasn't sure how to get this working on the other OS systems. What's flaky about this? What's the better electron way? I tried multiple solutions but I could only get this one working... :/ If you wanna take a crack at it, I'd be more than happy!

src/js/classes/app.js
FaultyFunctions: Hmm I don't remember why I needed to create a new one but I did try using the "yarnReady" one and ran into some issues. I'll try to update this and see what happens, maybe I just overlooked something.

src/js/classes/richTextFormatterHtml.js
FaultyFunctions: Huh I didn't know that was a thing! I just tested and yes it works!

electron/yarn-main.js
blurymind: This should be handled by electrons api, so you really shouldn't be limiting it just to windows users, even if you can't test it.
Keep the isDev check, but get rid of the win32 one.
If it doesn't work on Linux or mac, someone using those will submit a pr anyways

src/js/classes/app.js
blurymind: Hmm strange. What was the error? Why didn't you fix it there rather than creating a second event doing the same thing?

blurymind: Can't you detect when the dom is loaded from inside the electrons yarn-main.js file instead?

src/js/classes/app.js
```diff
@@ -90,6 +91,7 @@ export var App = function(name, version) {
   this.editingPath = ko.observable(null);
   this.$searchField = $('.search-field');
   this.isEditorInPreviewMode = false;
+  this.isEditorInPlayMode = false;
```
blurymind, line 94: I wonder if these should be in the ui class

src/js/classes/app.js
```diff
@@ -144,7 +147,13 @@ export var App = function(name, version) {
       return null;
     });
     window.addEventListener('DOMContentLoaded', e => {
-      this.data.loadAppStateFromLocalStorage();
+      if (self.isElectron) {
+        var event = new CustomEvent('yarnReadyToLoad');
+        event.app = this;
+        window.parent.dispatchEvent(event);
+      } else {
+        self.data.loadAppStateFromLocalStorage();
```
blurymind, line 155: This will completely break the feature of restoring app state from local storage on electron. You need to only prevent it when arguments are being passed to the executable. Please refactor this

src/js/classes/app.js
```diff
@@ -1190,26 +1233,27 @@ export var App = function(name, version) {
...
+
   // TODO: move to UI?
   this.togglePreviewMode = function(previewModeOverwrite) {
     const editor = $('.editor')[0];
     const editorPreviewer = $('#editor-preview')[0];
 
     self.isEditorInPreviewMode = previewModeOverwrite;
     if (previewModeOverwrite) {
-      self.togglePlayMode(false);
+      if (self.isEditorInPlayMode) {
```
blurymind, line 1253: can you move these ui related varianles to the ui class? isEditorInPlayMode,  isEditorInPreviewMode?

blurymind:
Left some review notes :) The PR will actually break the app restore state feature on electron, so that needs to be remedied for sure before it moves forward

FaultyFunctions:
Actually it doesn't break local storage on electron though. Check yarn-main.js. It has code specifically to load from local storage if no arguments are passed (a user didn't double click a file to open).

I just disabled it specifically in app.js because I didn't want it to load from storage and then after a second load from the file the user double clicked. Which was a bit jarring. So this way it just loads the file the user clicked right away and still uses local storage if no arguments are passed.

I'll check out your other comments soon!

electron/yarn-main.js
```diff
@@ -15,7 +18,19 @@ window.addEventListener("yarnReady", e => {
...
+  yarn = e;
+  // Open file on doubleclick on Windows
+  if (remote.process.platform === 'win32' && remote.process.argv.length >= 2 && !isDev) {
+    let filepath = remote.process.argv[1];
+    let filename = path.basename(filepath);
+    yarn.app.data.openFileOnStart(filepath, filename);
+  } else {
+    yarn.app.data.loadAppStateFromLocalStorage();
```
blurymind, line 29: ah I noticed it now - it would be much better if this conditional (if/else) split is handled in one location instead of two. Either here or in the data class

blurymind:
> Actually it doesn't break local storage on electron though. Check yarn-main.js. It has code specifically to load from local storage if no arguments are passed (a user didn't double click a file to open).
> 
> I just disabled it specifically in app.js because I didn't want it to load from storage and then after a second load from the file the user double clicked. Which was a bit jarring. So this way it just loads the file the user clicked right away and still uses local storage if no arguments are passed.
> 
> I'll check out your other comments soon!

Would it be possible to do the if/else in either yarn-main or in data? When you split the logic across two files it makes the code harder to maintain. The other thing is - can you in the feature do a separate pr for each fix? When you lump them up together like this, its very hard to catch regressions and harder to follow.Its easier for me to miss something when the diff becomes huge :)

blurymind:
> > > Also once this is merged and taken care of, I was wondering if we could get a new release so people can download the new changes? Thanks!
> > 
> > 
> > Releases stopped coming because the electron automatic builds stopped working. I welcome a pr to fix them. Maybe @daviddq can help?
> 
> I will take a look next weekend.

@daviddq thank you for the awesome and really quick fix 🎉 it worked like a charm. We now have fresh electron builds, straight out of the oven  🍞 😄

electron/yarn-main.js
blurymind: Ok so, instead of doing this in yarn-main, why dont you do it in electron/main.js? 
You can detect if dom is ready 
https://www.electronjs.org/docs/api/web-contents#event-dom-ready

then you can pass the args with
mainWindow.webContents.send(...

alternatively, why do you even have this here at all and not handle it all in data.js?
`if (app.electron?.remote?.process?.argv?.length >= 2)`

This yarn-main file is an interface for bundling yarn in alien electron aps. When we are handling our own electron code, I prefer to add to main.js

blurymind:
I reverted your implementation of the file loading from args and implemented a simpler approach (less loopy).

With that I think the PR is ok.

Please in the future dont bundle fixes in one PR.
Do a PR per fix - not several in one

PR #239, opened 2021-01-24
Fix build and release actions

daviddq:
I've replaced the "set-env" command used in the Actions since it's been tagged as "insecure" by GitHub.

Tested and working in my branch.


blurymind:
Thanks :) I will give this a try tomorrow

blurymind:
This works like a charm 😄 🎉 
@daviddq thank you!

Issue #248, opened 2021-02-28
Rename linked nodes in existing links?


s1-en:
On Twine, when I rename a node, it doesn't break already existing links but instead updates the links in the other nodes to reflect the changed identifier. Could this be implemented to Yarn Editor too?

FaultyFunctions:
This is currently functional but for some reason it's only coded to work with `[[<optional choice text here>|nodeName]]` syntax, not with just `[[nodeName]]`. So you need the `|` before the node name, then it will work.

This should be updated, thanks for bringing this up!

blurymind:
I think @daviddq originally implemented it. Maybe he could have a stab at this when he has some time?  :)

daviddq:
Hey! I remember coding this back in the day... and it kind of worked. Maybe there was some regression?

I'm completely disconnected with the process of this project, but I will try and download the latest code and make some tests. I promise nothing.

daviddq:
I prepared a quick PR to fix this: https://github.com/YarnSpinnerTool/YarnEditor/pull/274

blurymind:
Thank you @daviddq 
Merged now :tada:

PR #274, opened 2021-07-12
#248

daviddq:


blurymind:
This appears to cause a regression, instead of renaming the link it creates a new one
regressionRenameLinks

blurymind:
Thank you for working on this! Merged now :tada:

daviddq:
It works the other way around: renaming the node updates the links. You
probably saw it already since you merged the change.

blurymind:
Yeah I did. Sorry I totally forgot that detail. :)

Issue #272, opened 2021-06-30
Yarn Editor cant not save file


Yulovery:
Yarn editor can cretae a new file, save correctly. But when ever reopening file. Does not save in a form that unity or yarn reconize. It state its a yarn file, yet unity cant run it, and reopening it is just black space. As Yarn,or anything, just doesnt save as anything that work.

Please help me

blurymind:
fixed via https://github.com/YarnSpinnerTool/YarnEditor/commit/a7075bd956f88c76f14ed37d1b30d2333e2cfc5a

blurymind:
Can you please attach the broken yarn file? 

A before and after breaking version ideally

blurymind:
Please grab a new version if you are on electron
https://github.com/YarnSpinnerTool/YarnEditor/releases/tag/v0.4.163
see if this issue still persists

cvertee:
Looks like has some issues with saving again

1. Open editor and maybe git client for checking changes
2. Create new file with empty name OR skip this step and keep recently opened file
3. Open some other already created Yarn file
4. Edit it
5. Try pressing Ctrl + S, "Save" button - nothing works, but "Save As Yarn" is working

Although file is marked as saved in the bottom left corner, file isn't changing, only Save As helps

cvertee:
Reinstalling didn't help

OldDriverPickMeUp:
Hi, it happened that, this issue is blocking my work.

Had look through the code. In electron, yarn file can not be saved, because the iframe can not receive yarnReady event.

app.js line 353

`

    var event = new CustomEvent('yarnReady');
    event.document = document;
    event.data = data;
    event.app = this;
    window.parent.dispatchEvent(event); //this will fix the issue`

can some fix it and republish?

blurymind:
You could make a pr with the fix :) I will try to have a look tomorrow

OldDriverPickMeUp:
@blurymind hi, Todor.

Trying to make a pull request. but found a former commit 2 weeks ago.  

 ba7ee68: fix dispatching events to parent is problematic on electron

In this commit, 

in app.js line 353



   
`window.parent.dispatchEvent(event);`

was replaced with 

`window.dispatchEvent(event);`

I am quite confused now. When dispatch yarnReady event to top window, this will fix my issue. But i am not sure it will cause some other issues.

cvertee:
[...]

PR #276, opened 2021-07-25
#272 possible fix for save

cvertee:


blurymind:
This appears to fix the regression. Thank you :+1: :tada:

